PR #27, opened 2017-01-30
redux integration, packages update

mitjade:
Initial redux integration.

Had to add 'stage-1' to .babelrc, otherwise I kept getting:

> ERROR in ./src/store/reducers/counter_reducers.js
> Module build failed: SyntaxError: Unexpected token (9:8)
> 
>    7 |     case 'COUNTER_TEST_INCREMENT':
>    8 |       return {
>   9 |         ...state,
>      |         ^
>   10 |         test: state.test = state.test + action.value
>   11 |       };
>   12 | 

When I updated to latest apollo packages, you get a few `UNMET PEER DEPENDENCY` warnings:

> UNMET PEER DEPENDENCY apollo-client@0.8.1
> UNMET PEER DEPENDENCY graphql@0.9.1 invalid

But otherwise it is working OK.

Please review and make suggestions if something should be done in a different way.


vlasenko:
Thanks! I have only one concern about naming. Naming the redux counter as `test` might be confusing, people can think that this code has something to do with unit tests. I'd rather named it Redux counter or something like that

vlasenko:
Thanks a lot, Mitja! Great contribution!

PR #32, opened 2017-02-16
Offer Bootstrap 4 integration

josephdburdick:
Hi there, I'm thankful for this project and while integrating this for a current project requiring Bootstrap 4 I thought that I would give back and convert the existing structure to Bootstrap 4 alpha 6 and [Reactstrap](https://reactstrap.github.io), the replacement for React-Bootstrap. 

Bootstrap 4 *is* in alpha, yes, but it's extremely stable and, for those of us willing to develop front-end on the 'edge', I think it is a valuable offer for your project, at least in a separate branch, so people aren't deterred from using this due to the initial step of upgrading. React-Bootstrap says it's still working toward a version 1.0.0 release – and has been for several months – and by the time they release v1 for Bootstrap 3.3.7 I wonder if Bootstrap 4 will already be out of alpha.

That said, I completely understand if an alpha isn't a priority and want to contribute anyway. Thanks!


vlasenko:
I'm sold :) Many thanks for this contribution Joe!

PR #36, opened 2017-02-20
fix __DEV__ to be true when running tests

yishaic:
Fix https://github.com/sysgears/apollo-fullstack-starter-kit/issues/35


vlasenko:
Thank you!

Issue #38, opened 2017-02-20
Needs Autoprefixer


josephdburdick:
Autoprefixer needs to be added for CSS compatibility.

Issue #34, opened 2017-02-18
Merge SCSS files before injecting into HTML


josephdburdick:
Hi guys, it's me again :metal:. I noticed that Bootstrap SCSS and `styles.scss` are both included inside `html.jsx`. There's a few things here and I just wanted some clarification.
```
<style dangerouslySetInnerHTML={{ __html:
          require('../styles.scss')._getCss() +
          require('../bootstrap.scss')._getCss()
}}/>
```

1. The first thing that pops out to me is that `styles.scss` is loaded before bootstrap. I understand you might want to assign variables in `styles.scss` that `bootstrap.scss` would then use in place of their own, but any other css would likely be overridden by Bootstrap and that may be undesired for some developers.
2. You could leverage Sass imports and include `bootstrap.scss` internally at the top of the `styles.scss` file. If you're setting `styles.scss` to be before `bootstrap.scss` because you want to assign variables, a `_variables.scss` can be created. Note the underscore there, it allows variables to be scoped into the .scss file that imported it. Place bootstrap variable overrides in that and include as follows:

```
// styles.css
@import 'variables';
@import 'functions';
@import 'mixins';
@import 'bootstrap';

body {
  background: $gray-lightest;
}
```
Now my directory structure looks like this: (note: I added a styles directory because I plan on having quite a few modular files for Sass support)
```
style
├── _functions.scss
├── _mixins.scss
├── _variables.scss
├── bootstrap.scss
└── styles.scss
```
And now my `html.jsx` has the following: 
```
<style dangerouslySetInnerHTML={{ __html:
          require('../styles.scss')._getCss()
}}/>
```
No rush on replying but I am curious about this. Thanks guys!

vlasenko:
Hi Joe,

Yes, I think the order need to be changed vice versa, e.g. bootstrap.scss need to be loaded first, and styles.scss second.

The reason for separation is to speed up hot code reload for styles. Bootstrap styles are pretty big and we wanted to separate them out, so that they don't slow down reloading of custom styles.

josephdburdick:
Just following up here, it looks like splitting up the files like they are now increases build speed but one of the largest benefits of Bootstrap is being able to build with their variables available so developers can stick to a convention and not have to name things on their own. Running the project requesting a very popular variable returns in an  error:
```
// styles.scss
ERROR in ./~/css-loader!./~/sass-loader!./src/ui/styles.scss
Module build failed:
  background-color: $brand-primary;
[...]

PR #39, opened 2017-02-20
SCSS-related improvements

josephdburdick:
Hi, this PR includes Autoprefixer via PostCSS-loader inside Webpack (#38) and it also flips around the order of the SCSS files as discussed in #34. Let me know if you have any questions.


vlasenko:
Thank you Joe!

PR #40, opened 2017-02-20
Merge SCSS files. Fixes #34.

### LINKED ISSUE #34: Merge SCSS files before injecting into HTML

josephdburdick:
Let me know if you have any questions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "mocha": "^3.1.0",
     "mocha-webpack": "^0.7.0",
     "node-sass": "^4.2.0",
+    "postcss-loader": "^1.3.1",
     "raw-loader": "^0.5.1",
     "resolve-url-loader": "^1.6.0",
     "sass-loader": "^4.1.1",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "license": "MIT",
   "dependencies": {
     "aphrodite": "^1.0.0",
-    "apollo-client": "^0.8.4",
+    "apollo-client": "^0.8.6",
     "body-parser": "^1.15.2",
     "bootstrap": "^4.0.0-alpha.6",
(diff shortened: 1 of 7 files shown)


vlasenko:
No questions. Thanks again!

PR #69, opened 2017-03-10
Follow GCC naming convention for 'front/back'-end phrasing.

josephdburdick:
Hey guys I rephrased a few items in the README to follow closely to the [GCC naming convention](https://gcc.gnu.org/codingconventions.html#Spelling). Namely, using "front end"/"back end" for nouns and the hyphenated equivalent when acting as adjectives. Let me know if you have any questions.


vlasenko:
Nice, thank you Joe!

PR #73, opened 2017-03-12
Move styles into their own directory. Fixes #64

### LINKED ISSUE #64: Could we place style files in their own directory?
Hi guys. I create sizable, customized sites/apps where there's a lot of components and SCSS modules that are then imported in a large style index file (in our case, `styles.scss`). These files need to go somewhere and I like to keep things as organized as possible to improve scalability and so my style organization looks like this:
screen shot 2017-03-08 at 4 46 44 pm

The way things are structured now is fine if you're heavily reliant on Bootstrap and/or have minimal custom style definitions. If you customize SCSS variables or create a library of custom components that you want in their own individual files it's up to you to organize (and manually restructure every time you fetch from `upstream/master`.

My question is this: can we keep the .scss files in their own `style` directory? I can make a PR in the future if this idea is well-received but it probably won't happen in 1-2 weeks.

josephdburdick:
Hey guys, got around to doing this. If you have any questions let me know!


vlasenko:
Great, thank you Joe!

PR #75, opened 2017-03-13
Support fonts versioning and .woff

GiladShoham:
Support fonts versioning and .woff (in addition to .woff2)
Some fonts libraries like font-awesome use versioning for the fonts.
Improve the loaders to support it.


vlasenko:
@GiladShoham Looks good. Thank you!

PR #85, opened 2017-03-21
packages bump, replace var with let, rename inconsistent file names

mitjade:


vlasenko:
Cool, thank you very much Mitja!

mitjade:
Hi,

I know this starter kit is becoming a bit crowded with features and might be confusing to someone who is just getting started, but this is what attracted me to try it in the first place.

Example of count with subscriptions was really helpful to get me started with GraphQL and Apollo.

In my current project I would need a full CRUD functionality with subscriptions. For example something similar like posts and comments scenario.

I would like to make this first as an example with this starter kit and if anyone would find it useful it can be later added as an example like counter.

My goal is to have:
- cursor based pagination with infinite scroll for posts
- comments loaded with data-loader
- full CRUD for posts and comments
- subscriptions for both on create, update and delete

Would love to get some feedback on my implementation, of how to improve certain parts to make it more reusable and done in a proper way in case this is something you would find useful for this starter kit.

Regards,

Mitja

vlasenko:
@mitjade  Yes, I agree that its becoming crowded. But I would like to see all of the features of ApolloStack in this starter kit. I was thinking myself to add pagination at least as an example. But at the moment I'm still working on Webpack Persist GraphQL Plugin, it is non trivial and will have some compromises, still it should be very convenient during development.

So I would love to see the features you describe added to this kit!

Thanks,
Victor

josephdburdick:
I second the pagination​ and/or crud example idea as it would fit perfectly in this project as long as they were easily removable modules.

mitjade:
Should be the same amount of work as you have, when you remove counter example.

When I have something ready, will make a PR and we can take it from there.

PR #87, opened 2017-03-22
Pagination example for Relay-style cursor pagination and Infinite Scrolling using React-Virtualized

mitjade:
This is my first try of adding this examples to the starter kit.

Would love to get some feedback on how to improve it. I basically followed https://medium.com/@gethylgeorge/infinite-scrolling-in-react-using-apollo-and-react-virtualized-graphql-cursor-pagination-bf80617a8a1a#.60xx19fpm example and modified it for this starter kit.

A few things I already know could be done better, so if anyone would like to help I would really appreciate it.

- use something like https://github.com/domchristie/humps to transform database names like `post_id` to `postId` automatically.
- better `orderedFor` function maybe one that does no rely on `lodash` and move this code to some common utility file
- is Relay-style cursor pagination really need or could this be simplified, since params like `hasNextPage` are not even used
- I did not encode `after` cursor with base64 as GraphQL team suggests, I do not see any advantage?
- there is a bug if you scroll quickly that all posts do not get loaded and some get loaded twice (https://github.com/Gethyl/GraphQLPagination/issues/1)


vlasenko:
Hi Mitja,

Thanks for submitting! Two major points:
1. We cannot include this feature into starter kit until the bug with posts not loading or loading twice will be fixed.
2. It would be nice to have some posts generated during database seeding for testing purposes.

mitjade:
1. Will look into it soon, also posted an issue to the creator of the example this is based on.
2. Should already be done in https://github.com/sysgears/apollo-fullstack-starter-kit/pull/87/commits/de5d19b450c5e66ae57066236ee0c883e19fdcbe#diff-e605655f16f719cedeaf1c927b9c4169

mitjade:
Managed to fix the scroll bug by using `lodash` -> `unionBy`. Somehow `updateQuery` triggers multiple times in this case, so I just ignore duplicates when adding new edges.

vlasenko:
@mitjade Well, it can be seen as a temporary fix/workaround, but its certainly not good behavior and the root cause need to be identified and fixed.

mitjade:
`loadMoreRows` gets called multiple times with same variable in this case. Will see if I can prevent this for happening somehow. I presume it is a `react-virtualized` thing, so will see what can be done.

vlasenko:
@mitjade Is this `react-virtualized` dependency absolutely needed to make pagination example? Especially if it can be the cause of  issues?

mitjade:
It is not needed, but it adds a nice effect of loading while scrolling. I could just add a load more button ant the end of the list. Did some reading on `loadMoreRows` and it is expected that this is handled by developer, so I think I could find a way to make this work as it should.

vlasenko:
@mitjade If only thing `react-virtualized` does  - is nice visual effects - I would vote to not include it, since it's another dependency that people need to learn and it is not Apollo-thing, so its kinda offtopic

mitjade:
OK. Will remove it.

mitjade:
Now that I removed it, the code really looks more clear and I agree this should have priority before fancy features.

vlasenko:
Cool, thanks a lot, Mitja! From the brief look, I see only code styling issues, e.g. no space after `#` in graphql schema, while the rest of the kit uses space after `#`. Sometimes semicolons are used after imports, etc. So whenever you ready, I would like to merge this PR.

mitjade:
I think it should be ok now to merge.

I will now start working on mutations for post and comments and then make another PR for that.

vlasenko:
Cool, I appreciate your contributions very much. This one is lots of work, thank you so much Mitja!

mitjade:
A lot of years I was only observing the packages I used. Making fixes and improvements only locally. Only now I realised that by sharing this, in the long run, the one that benefits most out of this is probably me.

PR #95, opened 2017-03-28
Reorganise ui dir structure

mitjade:
Proposed reorganisation:

`ui -> post -> components`
`ui -> post -> containers`
`ui -> post -> graphql`
`ui -> post -> reducers`

`ui -> counter -> containers`
`ui -> counter -> graphql`
`ui -> counter -> reducers`

Deleted `store -> reducers`


mitjade:
@vlasenko Your thoughts on this.

Should we leave `ui -> components`, or move it in to something like `core` or `main`

vlasenko:
@mitjade I would prefer deeper decoupling with moving out UI and server-side code for the single feature into feature-specific dir, but shallow decoupling is okay too for the start. Maybe leave `ui ->  components` and `ui -> styles`  so that users understand they can write code there. Generally we should have some placeholder files for the users to quickly understand where they can start write their code. In addition to our examples.

vlasenko:
@mitjade But any way to decouple examples is fine. Let it be multi-step process. Let's start from decoupling approaches that can be implemented easily and then improve on that

mitjade:
@vlasenko So also extract schema and resolvers to separate files and into it's own dir?

For something like `server -> api -> middleware -> graphql.js` this might a bit more tricky.

mitjade:
@vlasenko Would this be than good enough to merge, or should we continue in this branch until we have complete decoupling.

vlasenko:
@mitjade Yes. I think examples should be fully decoupled ideally, including separation of schema and resolvers. Yes, it might be tricky to achieve this, but this is something to consider. If it's too difficult then let's do what we can and postpone hard things for now.

vlasenko:
@mitjade It's good enough to merge

mitjade:
Will try to extract schema and resolvers into separate files, since this might come in handy when you want to remove the examples, or as an example how this can be achieved.

Will attempt complete decoupling sometime later.

vlasenko:
@mitjade Yep, sure! Thanks!

mitjade:
@vlasenko Just give me the thumbs up if you agree with my proposal for schema and resolvers separation for post example.

mitjade:
I left count in the main schema and resolvers, so if anyone does not need to separate everything, we have both examples shown.

vlasenko:
Looks very good! I think it's good to merge this work, thanks!

PR #97, opened 2017-03-29
organise graph mutations: input

mitjade:
Just read [Designing GraphQL Mutations](https://dev-blog.apollodata.com/designing-graphql-mutations-e09de826ed97) and found it to be quite valuable information.

Not sure if I should follow all conventions, so for now I only introduced the `Input object`, since I thing it is really a good idea.

@vlasenko Do you think also `Input object` for just ID, `Unique payload types` and `Nesting` would be meaningful do add as well, or would this abstract the example too far and it would just confuse peolpe who want to get started with this.

### DIFF
--- a/src/server/api/graphqls/post_def.graphqls
+++ b/src/server/api/graphqls/post_def.graphqls
@@ -40,15 +40,36 @@ extend type Query {
 
 extend type Mutation {
   # Create new post
-  addPost(title: String!, content: String!): Post
+  addPost(input: AddPostInput!): Post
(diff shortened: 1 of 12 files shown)


vlasenko:
@mitjade Yeah, I think Input Object actually results in less code actually. So it's good. As for mutations with just ID, if Input Object results in less or the same amount of code then it's fine, if it results in more code - then lets not use it.

mitjade:
OK, so I will merge then as is.

PR #98, opened 2017-03-29
subscriptions on post comments

mitjade:
This is an initial draft how I am able to get CREATED, UPDATED, DELETED subscriptions working on post comments.

@vlasenko If you have some time, any feedback would be welcome.

Current TODO planed:
- move code from `subscriptions.js` into separate file, like for schema and resolvers
- figure out how to filter subscriptions if new comments are added (currently no filter is applied)
- retrieve proper `previousValue` for UPDATED and DELETED (currently not really used, except for delete id)
- make `mutation` field enum

### DIFF
--- a/src/server/api/graphqls/post_def.graphqls
+++ b/src/server/api/graphqls/post_def.graphqls
@@ -73,3 +73,13 @@ input EditCommentInput {
   id: ID!
   content: String!
 }
+
+extend type Subscription {
+  commentUpdated(commentIds: [ID!]): UpdateCommentPayload
+}
+
+type UpdateCommentPayload {
+  mutation: String!
+  node: Comment
+  previousValue: Comment
+}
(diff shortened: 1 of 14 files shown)


vlasenko:
@mitjade As a first iteration looks good to me. But feels like giving more thought to subscription logic might be needed to come up with more expressive solution.

mitjade:
Agreed. So for now, I will clean up a bit this approach and push. Later on we can improve on this.

vlasenko:
@mitjade Yes, thank you Mitja, sounds great!

mitjade:
Comment subscriptions are now filtered by `postId`, extracted subscriptions into separate file and removed unused `previousValue` and replaced it with `id` needed for `DELETE` mutation.

vlasenko:
Cool! Thank you Mitja!

PR #101, opened 2017-03-30
subscriptions on edit post

mitjade:
Added subscriptions on `editPost`.

My idea is not to have full subscriptions on all posts, but rather only on the one you are currently editing. I think making subscriptions work properly with pagination and possible sorting later on, might be quite a challenge. Also what if someone deletes a post you are editing, or just viewing. Also `pollInterval` every 10 seconds should do the trick of keeping the list up to date. This way you also showcase different solutions on how to update data.

Full subscriptions on comments I thing is great in case you would use comments as some sort of a chat.

Maybe I could add subscriptions on `num of comments` on all currently viewd posts, so this would be updated real time?

TODO:
- remember `Load more` variables, so if you have more than 10 records shown, polling does not reset this.

@vlasenko Do you agree with this approach?

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ predictable and buggy. To turn `React Hot Loader v3` on: set `reactHotLoader` fi
 
 - [GraphQL Cursor Pagination] Example of  [Relay-style cursor pagination]
 
-- [ReduxForm] Full CRUD funcionality in post example
+- Full CRUD funcionality with Subscriptions in post example, with [ReduxForm]
 
 ## Contributors
 
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   "author": "SysGears INC",
   "license": "MIT",
   "dependencies": {
-    "apollo-client": "^1.0.0-rc.8",
+    "apollo-client": "^1.0.0-rc.9",
     "body-parser": "^1.17.1",
     "bootstrap": "^4.0.0-alpha.6",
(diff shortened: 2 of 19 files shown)


mitjade:
Turns out `pollInterval` has it's own set of challenges, with load `fetchMore` call.

vlasenko:
@mitjade Not sure I follow. Just do what you think is reasonable and when I have enough time I will revisit post subscriptions code. I would vote for minimalistic showcase code, maybe have separate showcase for "pagination" and separate showcase for "posts and comments", if we have too many issues with making them work together at the same time. Just have more isolated and minimalistic showcases

mitjade:
@vlasenko Will investigate some more into this and if I can not find a reasonable solution, will split this into separate showcases as you suggest. My idea at the beginning was to have a fully working solution for Pagination, CRUD and subscriptions all in one, but I might bitten more than I can chew :)

vlasenko:
@mitjade Sure, sounds good!

mitjade:
Was able to make subscriptions work for posts with pagination as well. A few louse ends to tie up, but should be useful. Then over time this can be improved to be made better.

TODO:
- filter only on posts currently shown
- add subscription to update comment count in posts list

vlasenko:
@mitjade Sounds great!

mitjade:
Added filter for post subscriptions, removed comments from post list, to simplify the example a bit.

PR #118, opened 2017-04-10
Modularize starter kit

vlasenko:


mitjade:
Done some code style cleanup.

@vlasenko Please let me know if you agree with the code style. If not I can easily adjust this in my IDE and reformat files. I think it might be a good idea to do this now, since we are already making so many changes.

vlasenko:
@mitjade Yes, your code style looks good, thank you for helping with this! Could you review the code one more time? I did an attempt to make modular system more IDE friendly

mitjade:
@vlasenko I tried to create a new module with the minimum possible steps. It is really simple to accomplish this.

The only thing that seemed strange is that the new module was added before the `Posts` link, when using `addNavItems`. Maybe we would need a way to define where the link would be placed.

mitjade:
I would also like to add a short how to `remove`, `add` modules. Where should this be placed? Inside README.dm? Maybe start a Wiki?

mitjade:
`Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead` started to pop up. Should this be addressed in this PR, since lots of changes are already  made or wait till after this is merged and start a new PR?

josephdburdick:
@mitjade if the project was just updated to React v15.5 then yeah, PropTypes have been pull out of the main react package and now live on their own as the `prop-types` package.

mitjade:
@josephdburdick Yes, but did not update jet to v15.5? Just clean installed packages. In any case this will need to be taken care of soon.

vlasenko:
@mitjade About code style. I've spotted code like this: `process.argv[ 2 ]`, if it's your ide, please reconfigure it to do this instead: `process.argv[2]`

mitjade:
@vlasenko OK, will change this now. What about `prop-types`, should I change that as well?

vlasenko:
@mitjade Lets make code styling changes and other changes in separate PRs, because it is difficult to review the PR-related code. I'm going to rollback code styling changes from this PR.

vlasenko:
@mitjade Lets do `prop-types` change outside this PR, e.g. either commit to master or separate PR.

mitjade:
@vlasenko OK. Agreed

mitjade:
@vlasenko Any other changes planed or will this be merged soon now? I think it would be better for me to wait to reformat all files for code styling after this is merged.

vlasenko:
The `React.PropTypes` are accessed by `react-apollo`, `redux-form`, `reactstrap`

vlasenko:
@mitjade Okay, as you suggested earlier. Module import order might be imported. So I've decided to get rid of `webpack` magic with dynamical module loading in favor of doing explicit import in `modules/index.js`. And I have moved module connecting logic into `connector.js` to avoid circular imports. If you are fine with the latest version I'm going to merge this code to master, so that we move forward with other changes.

mitjade:
@vlasenko Quickly tried to add and remove a module and all works as expected. I am also in favour of less magic, even if you otherwise need a step more to do something.

vlasenko:
@mitjade Yes, I totally agree, thanks!

PR #121, opened 2017-04-11
disable persistGraphQL in DEV mode

zirho:
Thank you for the awesome project!

I just wanted to run queries on graphiql but got blocked by persistGraphQL on, even in DEV mode.
So I'm disabling it when it's DEV mode. 

Simply close this if you don't think it's a problem.

image


zirho:
I think I am missing something.

vlasenko:
@zirho Should be fixed via: https://github.com/sysgears/apollo-fullstack-starter-kit/commit/9f2be313a08f2a4211f68253d5865e7f30bcd762

PR #122, opened 2017-04-11
disable persistGraphQL in DEV mode

zirho:
this disables persistGraphQL in DEV mode, so that we can run queries on graphiql

### DIFF
--- a/src/client/main.jsx
+++ b/src/client/main.jsx
@@ -42,7 +42,7 @@ if (__CLIENT__) {
   );
 }
 
-if (settings.persistGraphQL) {
+if (!__DEV__ && settings.persistGraphQL) {
   // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved
   const queryMap = require('persisted_queries.json');
   networkInterface = addPersistedQueries(networkInterface, queryMap);
--- a/src/server/api_server.js
+++ b/src/server/api_server.js
@@ -28,7 +28,7 @@ app.use(bodyParser.json());
 app.use('/', express.static(settings.frontendBuildDir, { maxAge: '180 days' }));
 
 let queryMap = null;
-if (settings.persistGraphQL) {
+if (!__DEV__ && settings.persistGraphQL) {
   // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved
   queryMap = require('persisted_queries.json');
   const invertedMap = invert(queryMap);
--- a/src/server/middleware/website.jsx
+++ b/src/server/middleware/website.jsx
@@ -33,7 +33,7 @@ export default queryMap => (req, res, next) => {
         batchInterval: 20,
       });
 
-      if (settings.persistGraphQL) {
+      if (!__DEV__ && settings.persistGraphQL) {
         //console.log("backend queryMap:", queryMap);
         networkInterface = addPersistedQueries(networkInterface, queryMap);
       }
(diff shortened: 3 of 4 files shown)


vlasenko:
@zirho We certainly don't want to disable PersistGraphQL in DEV mode. You should be able to run queries using GraphiQL in DEV mode, without patching the kit.

zirho:
It doesnt seem to work without my patch. I think persist is for production by the purpose of it.

vlasenko:
@zirho In this starter kit persist is for development too by design. It is the purpose of persistgraphql-webpack-plugin to work both in development and production.

zirho:
I see, then how can I make graphiql work with query as well? for now, it's not working at least for me. Not sure how to make it work other than this patch.

vlasenko:
@zirho If you want to turn off persist graphiql in this starter kit no need for the patch. Set `persistGraphQL` to `package.json`. 

But graphiql should work with `persistGraphQL = true` too, it works for me with the latest version of this kit.

mitjade:
Works for me to, with:
```
{
  count {
    amount
  }
}
```
and
```
{
  post(id:1) {
    id
    title
  }
}
```

zirho:
@vlasenko Okay, so you are telling me 
While I am developing a project, I can turn it off in package.json. 
and when I am done, I should turn it back on and deploy it every time?
I don't think it make sense to me. Probably I'm missing something.

What I want is turn persist off only in dev mode.
So I can run queries in graphiql as @mitjade did here.

It does not seem to be working as is. Am I wrong?

mitjade:
@zirho What kind of an error do you get?

zirho:
image

http://localhost:3000/graphiql?query=%7B%0A%20%20post(id%3A%203)%20%7B%0A%09%09id%0A%20%20%20%20title%0A%20%20%20%20content%0A%20%20%20%20comments%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20content%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D

vlasenko:
The query above works for me in DEV mode with `persistGraphQL: true` in `package.json`
Please check that you have this code in your `src/server/api_server.js`:
https://github.com/sysgears/apollo-fullstack-starter-kit/blob/9f2be313a08f2a4211f68253d5865e7f30bcd762/src/server/api_server.js#L48-L50

zirho:
The code was not there so I added it and it works fine. 
Thank you guys. 
I figured that I just pull the latest right before you merge that change. 
Sorry about that I could have check it with the latest. 👍

vlasenko:
@zirho Sure, it happens. Thank you for confirming it works now.

Btw, using persisted GraphQL queries is not mandatory for production, so if you don't need them at all, just turn this option off. But if you want you application to use them, it will be nice, since this concept is rather new and needs testers.

zirho:
@vlasenko I like the persistgraphql stuff. They definitely will help the app scale better. 
Also that's the reason I found your project interesting and digging into it. 
Great stuff guys, I like it and hope I may be able to contribute in the future.

vlasenko:
Great to hear @zirho! 

We put much efforts and tried hard to integrate persistgraphql stuff well. Even webpack plugin have been written for this purpose:
https://github.com/sysgears/persistgraphql-webpack-plugin

I am glad that someone find it useful! Thank you!

PR #126, opened 2017-04-12
moved PropTypes from React

mitjade:
Updated some packages, including React to 15.5.

Removed PropTypes from React.

`Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.` is still there.

@vlasenko Is this because of some other packages using it?


vlasenko:
@mitjade Yes, I think some other package uses PropTypes

mitjade:
Just when I thought we got rid of one Warning, another pops up :(

mitjade:
@vlasenko Will wait with merging this until #118 is merged.

vlasenko:
@mitjade `React.PropTypes` is used by `react-apollo`, `redux-form`, `reactstrap`. Once all these packages migrate to `prop-types` we will get rid of this warning.

mitjade:
All three packages either have an issue, PR, or already merged the changes and awaiting release. So I guess it should not be that long.

vlasenko:
@mitjade Nice!

PR #141, opened 2017-04-14
reducer implementation initial try

mitjade:
This is a proposal how to implement client side cache update with `reducer` instead of `updateQuery`.

Just commented out `updateQuery `, so it can be compared what we could gain compared to `reducer`.

Currently only done on post comments.

### DIFF
--- a/src/client/modules/post/containers/post_comments.jsx
+++ b/src/client/modules/post/containers/post_comments.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { graphql, compose } from 'react-apollo';
-import update from 'immutability-helper';
+//import update from 'immutability-helper';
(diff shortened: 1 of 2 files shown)


mitjade:
@vlasenko So how do you feel about this approach, does it look any better? Will fix code later, should not use `return` in `switch -> case` :)

mitjade:
I guess I should try to unify mutation / subscription result, so `reducer` could be written more clearly and with less code.

The question is, what approach looks better, the `subscription` one with `mutation` name and `node`, or mutations. If mutations, then I would need to split `subscriptions`. I am leaning towards making mutations with same payload as subscriptions.

vlasenko:
@mitjade Mitja, should we close this for now and maybe reopen later?

mitjade:
Yes. I wanted to push some additional research on this front, but while doing this, realised that the changes I made add more complexity instead of simplifying things.

PR #188, opened 2017-05-18
Fix adding comment for Postgres DB

mairh:
- Fixing adding comment in case of `postgres` db.


vlasenko:
Fixes: #188

vlasenko:
Thank you Ujjwal!

PR #190, opened 2017-05-19
change graphql (post) edge type name not to be so generic

mairh:
- I ran into this problem when trying to add few other components in my app and had to spend some time debugging the issue. The problem is we are using `Edges` `PageInfo` as a `type` name for the post module, which is too generic and often mistaken by `edges` in the node query.
- I have added an explicit type name for post module to avoid confusion in the future.

FYI: http://stackoverflow.com/questions/33323894/why-are-edges-required-in-a-relay-graphql-connection (Also saw the explicit naming convention over here)


vlasenko:
Looks good to me, thank you!

PR #193, opened 2017-05-22
ES6 syntax refactoring

mairh:
- Using fat arrow function and class properties for binding methods.
- Reason for this change is `Function.prototype.bind()` methods are getting obsolete and ES6 "preferred" way is to use fat arrow syntax for private methods.


vlasenko:
Thank you Ujjwal!

PR #194, opened 2017-05-23
React helmet for SEO support

mairh:
- This PR should improve the SEO support for the starter kit.
- Supports server rendering
- Meta tags support based on nested routes
- Can be seen in action navigating to `counter`, `post list`, `edit post` and `add new post` routes

P.S - Would be nice to get some CR since I am not 100% sure if that's the optimal way to to go.


src/server/middleware/website.jsx
```diff
@@ -79,7 +81,7 @@ async function renderServerSide(req, res, queryMap) {
     delete apolloState.apollo.queries;
     delete apolloState.apollo.mutations;
 
-    const page = <Html content={html} state={apolloState} assetMap={assetMap} css={css}/>;
+    const page = <Html content={html} state={apolloState} assetMap={assetMap} css={css} head={helmet}/>;
```
vlasenko: `<Html>` should be updated in `renderClientSide` function as well in this file.

vlasenko:
@mairh Ujjwal, thank you for this contribution. It generally looks good except two points:
1. It should be supported by `renderClientSide` which is used when `package.json` -> `ssr: false`
2. I'm wondering why your implementation is not following Helmet docs for server usage:
https://github.com/nfl/react-helmet#server-usage

mairh:
@vlasenko Yep, I missed when SSR is false. I've added it to `renderClientSide` function too. Also, I've changed some param names and followed `helmet` docs as much as I could.

Could you please do another round of review now.

vlasenko:
Ujjwal, I have sent you an invitation to become Contributor of this project. This will let you commit directly to repository. You might still request Code Review when you think it is needed

mairh:
@vlasenko That's great, thanks! Yeah, I would probably won't push anything without a PR + CR if it is a new feature or major change.

PR #223, opened 2017-06-07
Add documentation about React Native support

mitjade:
This is just initial draft.

Please comment what is missing, not well explained or misleading.

@vlasenko Please add some `webpack`, `expo` and `haul` explanation.

@mairh Please add your comments as well, since you already started playing with this and can provide valuable info of what else is not clear, for someone who wants to start with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,8 @@ For running Android or iOS you need to set in package.json `ios` or `android` fi
 running both at the same time, since Expo does not support this.
 
 ### Running on a device
-You need to install [Expo] app on you Android or iOS device and then you can scan the QR, to start the app on your device. 
+You need to install [Expo] app on your Android or iOS device and then you can scan the QR shown in the terminal, 
+to start the app on your device. 
 
 ### Running in a simulator
 
--- a/README.md
+++ b/README.md
@@ -267,10 +267,14 @@ to start the app on your device.
 ### Running in a simulator
 
 #### Android
-You can use [Genymotion].
+You can use [Genymotion]. After downloading and installing you might need to install VirtualBox unless you already have it.
+Create a new emulator and start it. After starting the server Expo app should start on it's own.
+To bring up the developer menu press ⌘+M.
 
 #### iOS
-You need to install [xcode]. After the installation if you run this starter kit, simulator should start on it's own.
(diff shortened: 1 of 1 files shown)


vlasenko:
In the part about Haul, it is misleading. We don't use Haul for React Native packaging. We use Webpack for React Native packaging with the help partly from Haul parts, partly from React Native parts, partly from other packages.

mitjade:
Could you please rewrite that part, since I am not familiar how exactly that works.

vlasenko:
Sure

mitjade:
Maybe also add some mobile pictures to highlight that part as well?

vlasenko:
@mitjade Yes, it would be great to replace the video we currently have in a gif to give a feeling how working with the kit looks like

vlasenko:
I think we can integrate this PR in its current state

mairh:
Perhaps add a note to enable command line tools in `xcode` otherwise the simulator won't open. Whenever there is a fresh `xcode` install the command line tool dropdopwn is empty so you need to manually select the version from the dropdown list in order to make the emulator work.

screen shot 2017-06-07 at 11 39 36

mitjade:
Yes, thx. Forgot about that, since it's been a while since I've done that.

mitjade:
@mairh  Did you also need to install Watchman. I know I did when playing with react native, but am not sure if it is required here?

mairh:
Yes, watchman is needed for hot reload the changes to your device. It should be installed globally

vlasenko:
At the moment we don't support Hot Reload for React Native. Only Live Reload. Live Reload - fully reloads app on device when change is detected. Hot reload is other thing - it tries to hot patch the code of running application on the device without fully reloading it. We are going to support it later, its quite a big effort, but Live Reload should be enough though for practical needs with fast incremental builds and we will do them faster then now soon, I'm working on it

mitjade:
So Watchman is then currently not needed?

mitjade:
So I guess this is good enough for starters and we can improve upon this later on.

If you both agree, will then merge this PR?

vlasenko:
Yes, I agree

mairh:
👍

PR #232, opened 2017-06-12
Add module scaffolding support

vlasenko:


mairh:
I believe `cli` should in a separate repo. I don't clearly understand what is the point of having `cli` in the same repo as the kit since people will have to clone everything anyway.

The point of having `cli` is to bootstrap an app at the very basic without having the actual code in it (apart from templates)

Or maybe I am missing something in my understanding of `cli`??

mairh:
Nice one. I would add few more things in the `cli` templating

1) Add a basic unit test to show if the module is rendered correctly on mount

screen shot 2017-06-12 at 18 48 23

2) Since we are importing several packages in client/server and not using them. `eslint` will complain. So it's better to use them with default values.

screen shot 2017-06-12 at 18 49 06

3) Some comments on what each function is doing are probably also helpful.

Do you want me to fix the above things? Or do you think it's not necessary?

vlasenko:
@mairh Yep, nice catch. Would be great if you fix these please!

mairh:
Ok. I'll commit those changes

mitjade:
@mairh All contributions are welcome.

vlasenko:
Awesome, thank you Ujjwal!

mitjade:
@mairh Thank you!

README.md
```diff
@@ -97,6 +97,23 @@ for both web and mobile.
...
 planed in the future.
 
+### Module Scaffolding with CLI
+
+Basic Module Scaffolding is provided with the following command:
+```
+yarn cli module <module-name>
+```
+This will create all the necessary files to start developing on a new module. It creates `client` and `server` module.
+If you would like to ony add one or the other, add a second parameter like:
```
vlasenko: Typo `ony` -> `only`

README.md
```diff
@@ -230,6 +247,7 @@ While developing, you will probably rely mostly on `yarn watch`; however, there
 |`seed`|Seed sample database using SQLite. Use `--prod` flag to run in "production" mode.|
 |`migrate`|Migrate the sample database|
 |`rollback`|Rollback the sample database to previous state.|
+|`cli`|CLI tool, currently only used for Scaffolding modules.|
```
vlasenko: Should be `used for modules scaffolding`

README.md
```diff
@@ -97,6 +97,23 @@ for both web and mobile.
 Currently we do not yet support persisted queries. This can be used in this starter kit currently only for web, but it is
 planed in the future.
 
+### Module Scaffolding with CLI
+
+Basic Module Scaffolding is provided with the following command:
```
vlasenko: Maybe add some description about what is module, like:
"This starter kit encourages modular design of application features. Each feature should be designed as a decoupled module, deleting feature should ideally not break the remaining application. Basic feature module scaffolding is provided with the following command:"

mitjade:
Beat me to it :)

vlasenko:
@mitjade  Lol, I want this one to be merged, because I'm starting to work on lerna support

PR #234, opened 2017-06-14
addmodule, deletemodule cli commands

mitjade:
@vlasenko I renamed `client` `index.jsx` to `index.js`, to simplify the code needed to achieve this task.

Please comment if you have any reservations regarding this.


mitjade:
Will add the `deletemodule` also here and then I will merge.

mitjade:
add module into `index.js` for client and server with cli

renamed `module` to `addmodule`
added `deletemodule` for removing a module

PR #247, opened 2017-06-19
Knex support for other SQL DB's

mairh:
- Minor modification to support DB's other than SQLite3


vlasenko:
@mairh An idea to consider for the future is to enable database specific knex raw queries automatically based on the value of knex database driver used

mairh:
@vlasenko That would be the ideal solution. I will try to investigate it.

Issue #252, opened 2017-06-22
eslint


zirho:
Hi,

I got a bunch of eslint errors on eslint-vim integrated report.
I tried to find a fix for it, but I couldn't. 
It seems everything (.eslint*) is properly set up.

I don't get any errors with `yarn lint`.
I confirmed it's the latest master branch.

![image](https://user-images.githubusercontent.com/1251296/27418580-d8c4809a-56d0-11e7-9990-b1d576506707.png)

So I went check it out with Atom with eslint plugin, getting pretty same errors. 
![image](https://user-images.githubusercontent.com/1251296/27418618-0a31b88c-56d1-11e7-8643-ba9629f28d76.png)

Is anyone using editor eslint integration, not getting these errors?
Or any idea around this?

Thanks

zirho:
This is introduced by commit # ea13e75e34aef8b40309a39056ab28ad1d8d6612 
Trying to find what part of it now

zirho:
![image](https://user-images.githubusercontent.com/1251296/27419803-9d67386a-56d7-11e7-874c-11c6f0ef20a7.png)
This is the cause of it.

zirho:
Temporary fix is

![image](https://user-images.githubusercontent.com/1251296/27420357-7a9ec070-56da-11e7-8bd6-31d030211839.png)


![image](https://user-images.githubusercontent.com/1251296/27420366-85b068ba-56da-11e7-9626-c70488fb771d.png)


Ugly, but working.

zirho:
Thanks 👍

vlasenko:
@zirho I have pushed fix here:
https://github.com/sysgears/apollo-universal-starter-kit/commit/8353862bb6ef4e202234dbb4a82c0c93fb265c2f
Could you check if it works for you?

zirho:
That worked out gracefully
Thanks as always!

PR #253, opened 2017-06-22
add temporary fix for eslint editor integration

zirho:
I've made this PR just in case.
Please close if you think it's not appropriate.

#252


zirho:
That fix worked great!

vlasenko:
Thank you Youngjae!

PR #263, opened 2017-06-26
Refactor App.jsx as a Route

zirho:
I was trying to connect App.jsx with redux to get layout work with properties in the redux store.
The actual goal of mine was sidebar which can be opened and closed by redux actions from nested containers.

It didn't seem to work with any other approach.

Not sure this is the right fix for it.
Please look at this changes and give me an advice if there's better way to achieve this.

### DIFF
--- a/src/client/test-helpers/apollo_renderer.js
+++ b/src/client/test-helpers/apollo_renderer.js
@@ -1,6 +1,6 @@
 import ApolloClient, { addTypenameToDocument } from 'apollo-client';
 import { addApolloLogging } from 'apollo-logger';
-import { Router } from 'react-router-dom';
+import { Route, Router, Switch } from 'react-router-dom';
 import createHistory from 'history/createMemoryHistory';
 import { JSDOM } from 'jsdom';
 import { makeExecutableSchema, addMockFunctionsToSchema } from 'graphql-tools';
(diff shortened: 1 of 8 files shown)


zirho:
this make tests fail. I'll play around with this more.
Maybe I should try this way as well.
https://simonsmith.io/reusing-layouts-in-react-router-4/

mairh:
We can probably refactor the reusable components so that it is allowed to skip them in certain module routes. For e.g if you have a `sidebar` and `navbar` component you can probably skip them (don't render) in login / register / 404 pages. Conditional component rendering is a good use case for your PR

zirho:
@mairh Thank you for your input.
The problem that I was trying to solve is quite simple but quite hard to bring in this project.
I just wanted to get redux state in app.jsx which was very hard to keep other parts working as they are.
So I went a little far to get it working and ended up here with this PR.
If there's easier way to get redux store in app.jsx this will be an invalid PR.
Please let me know if you have any opinion regarding this. 
Thank you!

src/client/test-helpers/apollo_renderer.js
```diff
@@ -140,6 +140,10 @@ export default class Renderer {
   }
 
   mount(component) {
-    return mount(this.withApollo(<Router history={this.history}>{component}</Router>));
+    return mount(this.withApollo(<Router history={this.history}>
+      <Switch>
+        <Route path="/" component={component} />
+      </Switch>
```
vlasenko: Is wrapping into `Switch` needed here?

vlasenko:
@zirho Your Pull Request looks good, thank you!

src/client/test-helpers/apollo_renderer.js
zirho: It was not so I made another pull request with this.

PR #267, opened 2017-06-27
remove unnecessary Switch and div

zirho:


vlasenko:
Thank you Youngjae!

zirho:
Pleasure is all mine!

PR #292, opened 2017-07-13
Add each Route power to fully control page look

vlasenko:


src/client/app/page.jsx
```diff
@@ -15,12 +15,12 @@ const Footer = styled.footer`
   height: ${footerHeight};
 `;
 
-export default function App() {
+const Page = ({children, navBar}) => {
```
mairh: Maybe `Layout` is a better name?

vlasenko: @mairh Maybe.... I'm not sure

src/client/app/page.jsx
```diff
@@ -15,12 +15,12 @@ const Footer = styled.footer`
...
   return (
     <div>
```
mairh: Also, we can change `div` to `section` tag and remove all other `section` wrappers from individual component pages. So that we don't render another unnecessary `div`wrapper.

src/client/modules/page-not-found/containers/page-not-found.jsx
```diff
@@ -2,20 +2,23 @@ import React from 'react';
...
-        name: 'description',
-        content: 'Apollo Starter Kit - Page not found'
-      }]}/>
-    <h2>Page not found - 404</h2>
-    <Link to="/">
-      <Button className='home-link' color="primary">Go to Homepage</Button>
-    </Link>
-  </section>
+  <Page>
+    <section className="text-center mt-4 mb-4">
```
mairh: We can remove the `section` wrapper from all the component pages since `Page` is acting as a wrapper now.

vlasenko: Good idea, done

src/client/app/page.jsx
vlasenko: @mairh Renamed `Page` to `PageLayout`

mairh:
Looks good. Merge it whenever you feel like.

PR #304, opened 2017-07-26
Fix mobile build

mairh:
This fixed mobile rendering

screen shot 2017-07-26 at 14 24 30


vlasenko:
Oh, it was broken. Thank you Ujjwal for fixing it!

PR #312, opened 2017-08-01
Add custom scalar package `graphql-iso-date` as module

mitjade:
Initial commit on how to add custom scalar package into starter kit.

TODO: add example into post module


vlasenko:
Looks great, thank you Mitja!

PR #314, opened 2017-08-07
Use spin.js as a build infrastructure

vlasenko:
Refactor out all build infrastructure into a separate npm module:
https://github.com/sysgears/spin.js


vlasenko:
Guys, could you take a look at this PR and let me know your thoughts?

mitjade:
Will do so in the evening.

mairh:
Sure

mitjade:
I think this kind of infrastructure has both positive and negative sides.

Positive:
- it removes code so it would be easier to keep the project up to date
- easier to plug in different parts to get different setup

Negative:
- I have some changes done on this files in my projects that would then make it harder to change when in separate packages.
For example:
I had to switch from sass to less because, I use AntD.
I commented out automatic opening of a browser on load, since my default browser is not the same as the one I use for development.
I have a different template setup for scaffolding.

I have changes like this everywhere and if we start splitting up this kit into separate packages, we still need a way to be able to customise every part of it if needed.

mairh:
Where is the option to disable `iOS` or `android` build now when running `yarn watch`

mairh:
My 2 cents on @mitjade aforementioned points:

1) We can add support for less along with sass in the `webpack` by default using `less-loader`
2) We have to provide our own hard coded `webpack` stuff (like opening browser by default and so on) with some overridden settings in `.spinrc`
3) With scaffolding, you mean generating modules using `cli`? or what?

I really don't have any problem with splitting the build infra into a separate npm package. However, I wonder if people will have problems setting up their own infra needs without any proper way to override `spin.js` and how big job would that be to cater all those questions.

vlasenko:
@mairh @mitjade Thank you guys for your comments. The idea behind `spin.js` is that it should have config with as few typing as possible for simple cases and should be fully customizable for more advanced cases. At the moment it have functionality of the kit and this functionality will be growing, but customizability is somewhat limited - this will be improved as well.

1. How to disable iOS or Android. At the moment each node is a webpack instance. Node definition can have "features" field with list of features of the node, like "server", "web", "apollo", "react", these features describe the stack used for build node, for example ["apollo", "react", "es6", "android"] - is a stack used for "android" node, features, common to all build nodes  can be declared in "options" -> "features". Each node has also "roles" that as of now can be "build", "watch", "test" - the roles describe in what scenarios should the concrete build node used. So to disable android or ios, you can either declare empty roles field on that nodes or just remove these nodes declaration for `.spinrc`.
2. spin.js has some customization support already - you can override webpack config for each node in the file provided in `.spinrc -> options -> overridesConfig`. In that file you can override webpack config for each node. The export from this file should have the name of the node. You can also have `spin.config.js` file in the project root to fully control how the webpack config is generated.
3. As for `sass` vs `less`, I think `spin.js` should have out of the box support for `less` and `sass` feature, I will add it soon. I'm going to add `typescript` feature soon as well.

Generally we can integrate all your needs to `spin.js` in the form of out of the box supported options or features and then think how to make various customization approaches more generic. The idea is that `spin.js` should support all the possible JavaScript stacks and their customizations. It should be the tool that frees developers from the burden of writing build rules for JavaScript.

mairh:
Good explanation. I see the benefit of doing this. Thus people just have to focus on building apps rather spending time configuring or managing build infra.

I think a nice README.MD for spin.js repo is needed.

vlasenko:
@mairh Sure, I fully agree. I will add the initial Readme soon

vlasenko:
@mitjade Mitja, `spin.js@0.1.7` now supports both `sass` and `less` feature. To turn off automatic browser opening now one can add `openBrowser: false` option to the `web` build node. Do you have other modifications to build infra?

mitjade:
@vlasenko Thank you! Will check it out soon.

mitjade:
@vlasenko I finally gave this a try.

Less option works great.
How do I need to structure web node to include `openBrowser `?
I am able to not run `ios` and `android` by removing them from `builders`. But a lot of times I turn this `on` and `off`, so copy pasting it back is not as practical as before `true/false`.

I think all my other changes I've done in older projects are now already supported, but will check.

vlasenko:
@mitjade To have `openBrowser` in `web` node, it should look like this:
```json
{
  "builders": {
    "web": { 
         stack: "web",
         openBrowser: false
    },
  },
  ...
}
```

As for the second question, I think we can add `enabled` property to `builders` on the same level as `openBrowser`, which will be `true` if not provided.

mitjade:
This fails for me?

`SyntaxError: Unexpected token s in JSON at position 61`

```
{
  "builders": {
    "server": "server",
    "web": {
      stack: "web",
      openBrowser: false
    },
    "android": ["react-native", "android"],
    "ios": ["react-native", "ios"],
    "test": {
      "stack": "server",
      "roles": ["test"]
    }
  },
  "options": {
    "overridesConfig": "tools/webpack.app_config.js",
    "stack": ["apollo", "react", "es6", "less"],
    "backendUrl": "http://{ip}:8080/graphql",
    "ssr": true,
    "webpackDll": true,
    "reactHotLoader": false,
    "persistGraphQL": false,
    "frontendRefreshOnBackendChange": true
  }
}
```

mitjade:
stack needs to be in quotes, fond the problem

vlasenko:
@mitjade Oops, yes, in JSON keys need to be double-quoted, sorry

mitjade:
@vlasenko Works!

vlasenko:
@mitjade Added support for `builder.enabled` property into `spinjs@0.1.10` and updated this PR with example in `.spinrc`

mitjade:
@vlasenko Tried it out and it works great!
I can already see the advantages of doing things like this in separate packages. It was much easier just to bump `spinrc` version and change one file, instead of up-merging. Especially when you have multiple projects, running on different configs and not recently updated.

mitjade:
I always had at least conflicts on my `package.json` and `yarn.lock` file if noting else.

vlasenko:
@mitjade Yes, thats the idea - to free developers from writing and maintaining complex JavaScript build rules. There are many challenges to actually achieve this, but I think its worth it.

vlasenko:
@mitjade I don't like `yarn.lock` at all. I think it creates more problems than it solves. I resisted to add it to this repo for a long time, then added to see how life goes with it and I don't like it. I need to merge changes far too more often then before, I have issues with greenkeeper and I don't think that locking out dependency package versions is of big help.

mitjade:
I would rather just remove all `^` in `package.json` file and have no lock file and then update packages when needed for every version.

vlasenko:
@mitjade Yes, sounds good to me.

mitjade:
@vlasenko Not sure that would help much. I do not think you can get rid of `yarn.lock` and it is recommended to be included in source control [(yarn-lock)](https://yarnpkg.com/lang/en/docs/yarn-lock/)

vlasenko:
@mitjade Yes, I know, I have read articles about `yarn.lock`. Despite the rationale in these articles ,periodically, I have a desire to throw `yarn.lock` away :) But I haven't come to the intolerable point yet, so `yarn.lock` lives in the repo for now

vlasenko:
Initial plugin architecture has been implemented for Webpack config generation in `spin.js` and all technologies supported have been moved into separate plugins. I'm working now on Webpack config execution plugin architecture.

PR #316, opened 2017-08-17
upgrade packages

mairh:
- We had quite a few old package. I have updated some which weren't causing any breaking changes.
- Note: `graphql-server-express` has been renamed to `apollo-server-express` so I have replaced the old package with the new one.

Before
screen shot 2017-08-17 at 12 11 11

After
screen shot 2017-08-17 at 13 28 18


mairh:
I'll do a thorough testing again and merge if everything is in place.

PR #322, opened 2017-08-24
Post module refactor

mitjade:
on post add remain on page

### DIFF
--- a/src/client/modules/post/components/post_edit_show.jsx
+++ b/src/client/modules/post/components/post_edit_show.jsx
@@ -15,7 +15,14 @@ const onSubmit = (post, addPost, editPost) => (values) => {
 };
 
 const PostEditShow = ({ loading, post, navigation, subscribeToMore, addPost, editPost }) => {
-  if (loading) {
+  let postObj = post;
+
+  // if new post was just added read it from router
+  if (!postObj && navigation.state) {
+    postObj = navigation.state.params.post;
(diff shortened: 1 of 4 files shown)


mitjade:
@mairh I managed to combine post add and edit page and create a transition without blinking, for both  web and mobile. Could you take a look, if this is something you intended when we ware discussing this a while back.

I still need to implement nested mutations, so you would be able to add comments before first save, but maybe this is already better to what we have now.

I also plan to add two buttons Save and Save and list to be able to skip one click in case you want to also go back to list..

mairh:
I will have a look tomorrow morning. Thanks for the PR

vlasenko:
@mairh Are you still going to take a look or should we merge this one?

mairh:
I apologize I was too busy to look into it. I will surely review it today and merge it.

mitjade:
@vlasenko I still plan to do some work on this refactoring post example, after I'm done with auth, but in it's current state can already be merged if anyone would find it useful.

mairh:
It looks good. I will merge it

We can take the next 2 points later on when you are done with auth stuff

> Quote 1: I still need to implement nested mutations, so you would be able to add comments before first save, but maybe this is already better to what we have now.

> Quote 2: I also plan to add two buttons Save and Save and list to be able to skip one click in case you want to also go back to list.

PR #325, opened 2017-08-31
Auth

mitjade:
Initial auth implementation.

Still work in progress.


vlasenko:
@mitjade Mitja, could you describe security concerns with Authonentication implemented in this Pull Request. Why are you using JWT tokens, when the token is generated, what payload does it has. When refresh tokens are generated and why are they needed? I think these concerns need to be discussed and addressed before merging this PR into the master

mitjade:
@vlasenko To be honest I do not really have a lot of experience on this field. If I want good protection I use Server and Client side certificates and that's it. But this can not work for may cases, because issuing new client certificates for for example online store, will not work out :)

After some googling JWT kept popping up so I decided to give it a try, since many examples on how to achieve this ware available. So I am not saying this is the best approach, it is the one I was able to make work and am up for suggestions on how to change it or make it better.

Here is how it works now:

I use 3 tokens and store them in cookies.
1) `x-token` with payload `id, username, isAdmin` that expires every minute
2) `r-token` that is same as `x-token` but with `httpOnly` to prevent client side tampering
3) `x-refresh-token` with payload `id` that expires every 7 days
4) `r-refresh-token` that is same as `x-refresh-token` but with `httpOnly` to prevent client side tampering

Whenever a login mutation is made and user with the appropriate password exists, this tokens are created and sent back to client as result. Tokens are created based on `SECRET` and for `x-refresh-token` also user password in used in combination with `SECRET`, so the token can be denied if password has changed. Client then stores them in `localStorage` and sends them back to server through `headers` on every `request` or as `connectionParams` with `subscription` connection.

On the server if `token` is available in `header` or `cookie` it is verified and checked if expired. Id `x-token` has expired new one is created based on `x-refresh-token`. If token is successfully verified `user` is added to the graphql `context` and `cookies` are saved.

On logout `localStorage` and cookies are deleted and on first server request, also `httpOnly` cookies are deleted if header has no `x-token`.

Why use tokens and refresh tokens? Reduce the amount of times you need to query database. This way you only need to query database when `x-token`needs to be initially created or is expired and depending on your requirements this can be shortened or prolonged. 

Why use cookies? This was the only way I know how to make it work with SSR, plus you can get additional security with `httpOnly` cookies, that can not be changed client side. This way is someone changes cookie, you can compare `x-token` and `r-token` and if they do not match, revoke access.

Why use `localStorage`? Am asking my self the same question. Initially the example I followed was done with this approach, but I then realised SSR is not working and you can easily change it to login as some one else in case you are able to get someones valid token. But if client has cookies disabled this, could be a fallback option, so the login still works is someone does not refresh site.

Comments, concerns, recommendations are welcomed.

vlasenko:
@mitjade Thank you for explaining this. Your approach looks good to me. I'm fine with merging this PR sooner rather then later

mairh:
I know this PR is still in progress, but it would be nice to have the following things added before we go ahead and merge this PR:

1) Client and Server tests for user module.
2) Basic validation - In case email / username already exist.

bastianwegge:
I'd love to help on this one. Are you appreciating help @mitjade and which timezone are you in?

mitjade:
@bastianwegge Thank you for offering! Would appreciate any help. I'm from Europe.

Todos:
- improve security
- mobile integration
- write tests
- write documentation
- improve validation (register, login)
- email confirm, lost password
- edit user / users
- improve / extend authorisation (roles)
- move all logic to user module (server side), so everything can be easily removed if not needed

Pick anyone you would need for your project, like to work on, on feel comfortable doing. Just let me know before you start, so we do not work on the same thing.

I will write here as well before I start something.

mitjade:
@vlasenko I accidentally merged this into master, could you remove this from master. Do not know how to do this the right way.

mitjade:
@vlasenko I think I managed to revert. Please check if all ok.

bastianwegge:
@mitjade I'd love to discuss your implementation first. As I don't want to step on your feet I think it makes sense to fully understand what you did first. I'm on the train for nearly the whole day, but I'll try to look through your implementation. Do you use gitter or slack?

mitjade:
@bastianwegge gitter will do.

mitjade:
This PR got closed when I accidentally merged this into master. Please refer to https://github.com/sysgears/apollo-universal-starter-kit/pull/332

PR #332, opened 2017-09-07
Auth

mitjade:
Fresh PR for Auth. Old one can be found [here](https://github.com/sysgears/apollo-universal-starter-kit/pull/325).


mitjade:
@vlasenko I would like for this kit to work even if you delete the `user` example.

Currently I have changes in `server/api_server.js` and `server/api/subscriptions.js` that prevent this to work. If could move this changes to user module, this would work.

I have similar changes in my other projects where I use ether `mqtt` or `udp`. Could you help me out with this?

Currently I do not know how to do a `app.use((req, resp, next) => {...});` inside a module and how to have `req` and `res` available in `server/api/subscriptions.js`?

vlasenko:
@mitjade Sure, I can help with this

bastianwegge:
@vlasenko what's currently holding your `lerna` branch back? A lot of the work on this branch doesn't take the `lerna` branch into account and will definitely crash as soon as we're merging. So maybe coming up with a strategy would make sense.

I also would suggest using smaller PRs rather than PRs that are open for longer than a week or two. What do you guys think?

vlasenko:
@bastianwegge `lerna` branch will have `webpack` instances running in different `node` process, they need to communicate between each other. We need to have infrastructure in `spin.js` and webpack plugins we depend on to be ready for that. Another cons of lerna is that it takes much more time and space to install dependencies via npm or yarn as compared to single module project. And package hoisting introduced by lerna is not really working good with Webpack and its plugins. So lerna branch really have tons of problems to solve. But we definitely want to work in this direction.As for `crash` of `lerna`, branch it is not really the case, the `lerna` branch depends only on our build infrastructure, which is insde `spin.js` now. Our PRs with new application level feature is irrelevant to lerna branch

vlasenko:
@mitjade I have looked at your code. To make it 100% modular, I think two things need to be done:
1. All context creation should still stay in `module.createContext`. Your new code in `subscriptions.js` should be moved to user module createContext, but this createContext function need to accept parameter - connectionParams, this way you will be able to move that code there.
2. `app.use((req, resp, next) => {...});` should be moved to modules. Module should have possibility to export array of middlewares and inside `api_server.js`, we should loop over all modules middlewares and execute app.use(middleware) for each item.

Does this sounds good?

mitjade:
@vlasenko Exactly! Would you be willing to set this up. I can move the code for `user` example my self, I just need the two above mentioned thing setup.

mitjade:
I think this will be handy in many other scenarios and make this kit even more modular.

vlasenko:
@mitjade Sure, will do now

mitjade:
Maybe even extract some of the current implementations we have now in `api_server.js`.

vlasenko:
@mitjade Yes, we can make the kit much more modular. But lets work on one problem at a time :)

mitjade:
@vlasenko Of course. Always getting ahead of my self :)

vlasenko:
@mitjade Done, please take a look

mitjade:
@vlasenko Checking now!

mitjade:
@vlasenko Looks great! Works as before and now if I remove `user` module, kit still works.

Will take me a while to figure out what you did. This is really awesome. With this approach I can clean up my other projects and make them a lot easier to up-merge.

This kit is for sure moving in the right direction! You are the man @vlasenko!

vlasenko:
@mitjade Good to know, thank you so much Mitja!

mitjade:
@vlasenko I would like to ask for just one more favour if I may.

I would actually like to create cookies inside a login resolver, because now I create it in the middleware. If someone would to set his `localStorage` with a valid token before they login, he would still be able to login, since I only create the `cookie` after I get the first response back from the client and would really like to resolve this.

But again do not know how to pass the `req` inside `subscriptions.js`. There you set the first parameter to `null` in `createContext`. If you know how ta access it there, I can add `req` to `context` and be able to access it in my resolvers.

Or if there is another way to create a `cookie` inside a resolver without having access to `req`, since I do not think passing it to `context` is a good idea anyway?

vlasenko:
@mitjade In `subscriptions.js` the work is done with WebSocket connection, not HTTP connection. So you need WebSocket-specific mechanisms to simulate working with Cookies. You need to figure out what is available for WebSockets as a replacement for Cookies.

mitjade:
@vlasenko OK, thx for the info, will look into it.

vlasenko:
@mitjade I think to support auth concerns over websocket, we should go a bit lower level and implement auth concerns on top of:
https://github.com/websockets/ws
and then pass this extended with auth WebSocket to `subscription-transport-ws`

vlasenko:
@mitjade Actually `subscriptions-transport-ws` has a way to support refresh tokens (so need to go to lower level of websockets):
https://github.com/apollographql/graphql-subscriptions/issues/68

mitjade:
@vlasenko Yes. I am already ding this. It is only a matter of when to set the cookie.

I figured out that at the time of login mutation `req` is set, so I just check if available and then set cookies.

vlasenko:
@mitjade How do you support refreshing tokens over WebSockets ?

mitjade:
@vlasenko I think so.

mitjade:
I pass it in with `connectionParams` and then in `createContextFunc` check if token is valid else refresh. This should now be the same as http.

mitjade:
@vlasenko Pushed the changes. Hopefully someone who knows a bit more then I about security can answer if this approach is ok. I intend to keep an eye on this and improve it if I find something can be done better on more secure.

vlasenko:
@mitjade The problem is that Auth should be able to work over WebSockets without relying on HTTP at all. Because in the future it is quite possible that Apollo will switch to WebSocket-connection only. Is this the case now?

mitjade:
@vlasenko I do not think so. I set `connectionParams` for `subscriptions` in `client/app/main.jsx` (line: 58). Do not know how this could be set without HTTP. I guess this is a bit beyond my current comprehension.

vlasenko:
@mitjade Okay, lets reiterate over security one more time. What is current security model of the authentication process?

mitjade:
1) Login mutation (server) -> tokens cereated and storred in `cookies`.
2) Login mutation result (client) -> tokens stores in `localStorage` and passed to `header` and `connectionParams` for `subscriptions`
3) On every request (server) -> tokens are read from `header` or `cookies ` and then verified and refreshed if needed -> if all ok `user` is added to `context`
4) Resolvers that need auth (server) -> check if `context.user` exist
5) Routes that need auth (client) -> check if tokens exist in `localStorage` and verify tokens before allowing access to route.
6) On logout `localStorage` and `cookies` are deleted (client) and on first request (server) `httpOnly` `cookies` are deleted as well.

vlasenko:
@mitjade I would imagine the security model with token authentication could be the following:
1. Server issues and signs with its secret key a long-term (7 days) authentication token and sets it in httpOnly cookie (we can still do that even on WebSocket connection, I believe). HttpOnly cookie gives us the guarantee that frontend cannot change this cookie in any way and cannot impersonate valid long term authentication token of the other user. Server need to issue new long-term authentication token, periodically,  when new WebSocket / HTTP connection is established, so that effective period of 7 days stay intact
2. Server issues and signs with its secret key a short-term (1 minute) token and passes it to frontend.
3. Each time the frontend needs to do authentication and authorization of the user, it checks short-time token. If short-time token is invalid it asks for new short-time from the server in order to proceed and server issues new short-time token.

Does this describes current security model?

mitjade:
@vlasenko Not quite. I also send refresh tokens to client, in case `cookies` are disabled and only `localStorage` is available. But we could just remove `localStorage` option, since it is a security risk anyway and simply not allow login without `cookies`.

mitjade:
Someone could disable then cookies just for the pursue of trying to bypass cookie validation.

vlasenko:
@mitjade Storing long-term token in LocalStorage has the risk of leaking out, but it is not directly a flaw in security. It is a problem only if we allow leakage of information from local storage.

mitjade:
Long term tokens also currently do not get refreshed and will expire after 7 day, or if a user shanges his password.

vlasenko:
@mitjade Long term tokens should be refreshed too. Otherwise login sessions will terminate unexpectedly for the users

mitjade:
@vlasenko I'm afraid I am not following you any more regarding `LocalStorage` leaking out?

Also we need to take in to consideration how this will look on mobile and use of `AsyncStorage`.

vlasenko:
@mitjade I think you should send request to the server when short-time token is expired, I don't see where you do it?

vlasenko:
@mitjade Using local storage is not a problem I think, but perhaps we need to take some steps to prevent leaking info from local storage

mitjade:
@vlasenko Tokens are send every time in  `applyBatchMiddleware` -> `client/app/main.jsx`

vlasenko:
@mitjade I understand this, but if user is active only in frontend and this doesn't result in requests to the server, you MUST send request to the server separately to refresh expired token

mitjade:
@vlasenko This is not taken care of. This should be done with `subscriptions`?

mitjade:
Now I understand. Client `auth` then stops working after 1 minute.

vlasenko:
@mitjade Yes. Yes better do with subscriptions, because Apollo is heading to WebSockets

vlasenko:
@mitjade Sorry, actually, it doesn't make a difference. It is enough to send any request to the server over GraphQL, doesn't matter what is this request

mitjade:
@vlasenko So I need to have a 1m ping request. Where could I attach this?

vlasenko:
@mitjade Now for long-term token. On the server we should pick only one long-term token from HTTP/WS connection. We prefer to pick it from Cookies if it exists there, otherwise if it does not - we pick it up from the headers. There is no reason to compare them together from security perspective. There is no reason to pass from frontend to the server short-time token at all. Short-time token is only needed on frontend

vlasenko:
> @vlasenko So I need to have a 1m ping request. Where could I attach this?
@mitjade No you don't need 1m ping request. WHEN short-time token is invalid on client you should send request to the server in order to get new short-time token. This will happen when you check short-time token.

mitjade:
@vlasenko If I do not pass long-term token to the browser, I do not know how to then pass it to the `WebSockets `. I can read `cookies` from WebSockets, but again if disabled this will not work without `LocalStorage` and I know how to pass this to WebSockets only if it is first passed it to the browser.

And I actually use long-term token to check client side if token has expired, so ping in this case is not necessary, but I agree not as secure.

Also we then have the problem of waiting for the server response before we can authenticate the token client side after it expired.

mitjade:
So the question here is, do we have to check client side every minute, or we leave the user being able to browse client data even if shot term token has expired. And only after first server request refresh it.

vlasenko:
@mitjade When we need to check user rights in UI and we determine that short term token has already expired - we send the request to the server to get back fresh short term token and wait for server response.  

In addition to that we can also send request to the server with the purpose of getting fresh short term token, asynchronously, ahead of time, say 5 secs to expiration time. But this is additional measures to make UI not laggy.

vlasenko:
@mitjade Why you need long-term token on client side? It should only be needed on server side

mitjade:
@vlasenko As I mentioned, I do not know how to pass it to `subscriptions` otherwise.

mitjade:
@vlasenko I know there are many things we still need to improve in auth example, but I only have today and tomorrow left, then I got for a short vacation and then I start on a new project and will not have much time to work with this.

Please specify priorities, what needs to be addressed before this is merged, since I do not want this PR to get stuck at 90% done and never merged.

All that you are saying is absolutely true, but I might not be able to get this working, since I do not exactly know how to do this. I can easily write some docs, take care of login, registration validation concerns and similar.

vlasenko:
@mitjade Why you need long term token in subscriptions? You need long term token on the client side if and only if you have no cookies support and using local storage. In this case you obviously need to store this token on client, since there is no other place to keep it.

vlasenko:
@mitjade Okay, let me handle security side of the things, token transport, etc. Do everything else. Error handling, platform support, tests, etc.

vlasenko:
@mitjade The only aspect is I'm worried now is that token generation and consumption mechanics. This is holding me off from merging. But if you have too little time, its fine, I will take care of this stuff. The problem is that properly implementing this stuff will take much time, I think.

mitjade:
@vlasenko That would be best. I'm sorry for putting this on you as well, but as mentioned not really my area. I wanted to do this because I think it is a key feature to be able to have when selecting what starter kit to use and many people have requested it. Every body liked meteor because it came with this out of the box.

Yes I underestimated the amount of work needed to make this work correctly.

vlasenko:
@mitjade Yes, I wanted to use something ready-made for the auth. Didn't really want to implement it from scratch, all these tokens and refresh tokens mechanics and UI stuff, way much efforts needed to have this all. But if there is nothing decent exists, then we have no choice.

mitjade:
[This](https://www.youtube.com/watch?v=agKRMEjqb5M&feature=em-uploademail) just came out. Will check it out if anything useful is shown.

mitjade:
@vlasenko Maybe @bastianwegge can help out with this, since he offered to help.

@bastianwegge Could you tackle any of the concerns @vlasenko has pointed out?

bastianwegge:
I've looked through most of the code by now. I'll be looking at this discussion later today. (But) As I said before it makes sense to do small steps rather than big PRs.

bastianwegge:
I'm wondering a bit about the implementation details and also over the discussion here since there seems to be a lack of knowledge when it comes to security. Is there a reason not to use something like https://github.com/jaredhanson/passport on the server? I mean there are over 300 different strategies to do authentication (web-app/mobile-app or intranet/internet or OAuth/OAuth2) and passportjs is backed by experts in that fields.

I'm impressed by the current state of this branch, this is really good work! As you're short on time @mitjade maybe it makes sense to do some polishing and finish off before going on vacation.

@vlasenko I'm willing to take care of further developing the state of authentication on this repo and maybe chat with you about the server-side. Would you then - after thoroughly reviewing - be willing to merge this PR?

I'll put in some more time now (like an hour) but I don't really want to touch without fully understanding your intentions. I'm in gitter if one of you has got time 😀

vlasenko:
@bastianwegge I will be glad if you will take care of this PR, while Mitja is on vacation. I'm all for using passportjs, instead of inventing our own Authentication protocol.

mitjade:
@bastianwegge The problem I had was that all existing libraries ware not written with `graphql` in mind. I struggled to get it to play nice with the way this kit is setup, so tried a different approach and kinda made it work.

mitjade:
@bastianwegge Tried to do some cleanup and am just no making much progress, so I'm just going to stop at this point and leave it up to you, since you know much more on how to handle this in the correct way.

Current things I know, that are not working ok:
- On logout, when `client.resetStore()` is called, there it a [bug](https://github.com/apollographql/apollo-client/issues/1186) in `apollo-client` and you get stuck on loading
- If you login and wait for a minute for the short-term cookie to expire and then do a page refresh, current implementation does not work, if some kind of token refreshing is going to be made to keep the short-term cookie alive, this will not be a problem

mitjade:
I will try just try to add some docs, tests and some small changes to login, register, profile and users pages.

bastianwegge:
@mitjade as mentioned before, I will not "take over" the PR. I'm willing to help with small PRs improving your solution after it's merged in `master`. Discussions on big PRs are almost always more philosophical than technical.

@vlasenko can you review and merge the PR or assign some points that seem undone for you so @mitjade can "finish" this?

vlasenko:
@bastianwegge I will merge this PR, when security concerns raised in:
https://github.com/sysgears/apollo-universal-starter-kit/pull/332#issuecomment-327775343
will be addressed.

mitjade:
I fixed the second problem regarding short-term cookie being deleted and some other mistakes I made.

mitjade:
@vlasenko The way this PR is now, is what I am able to accomplish with my current knowledge. I agree it is far from perfect, but it works and might be useful for people to play with and have something working. If you think this is needs to be improved before merged, then I kindly ask for you to take a look and improve it before merging it.

@bastianwegge If you are willing to take a look at the current implementation and suggest some small improvements, I can still look into it today and improve it a bit my self.

src/client/app/main.jsx
```diff
@@ -19,15 +20,45 @@ import '../styles/styles.scss';
...
 });
 if (__CLIENT__) {
+  networkInterface.use([{
+    applyBatchMiddleware(req, next) {
+      if (!req.options.headers) {
+        req.options.headers = {};
+      }
+
+      req.options.headers['x-token'] = window.localStorage.getItem('token');
+      req.options.headers['x-refresh-token'] = window.localStorage.getItem('refreshToken');
```
bastianwegge: Is there a way to abstract this into the user-module?

vlasenko:
@mitjade Okay, lets merge it and continue from there

src/client/modules/user/containers/auth.js
```diff
@@ -0,0 +1,156 @@
...
+
+          await client.resetStore();
+
+          window.localStorage.setItem('token', null);
+          window.localStorage.setItem('refreshToken', null);
+
+          if (history) {
+            return history.push('/');
+          }
+          else if (navigation) {
```
bastianwegge: unnecessary else

mitjade: agreed

src/client/app/main.jsx
mitjade: will take a look

mitjade: Something like this? https://github.com/sysgears/apollo-universal-starter-kit/pull/338
Still working on connectionParams

mitjade: should be done

src/client/modules/user/containers/auth.js
mitjade: https://github.com/sysgears/apollo-universal-starter-kit/commit/a54b89e7534f8c754a15e45cdf127593c493c8bd

PR #336, opened 2017-09-07
add nvmrc

bastianwegge:
`.nvmrc` allows for automatic switching to a specific Node.js Version if you're working on more projects.


vlasenko:
I don't think we should lock Node.js version used for the starter kit via `.nvmrc`. I think locking should be done by the people after cloning/forking the kit, if they need to.

bastianwegge:
Why is this locking? You already defined a Node.js Version inside of the package.json that will bark out loud when you're not on that specific version.

vlasenko:
@bastianwegge Until we have upgraded to Yarn `1.0.1`. It didn't bark out loud and engines section were meant only for Heroku. After I have merged your PR and updated to Yarn 1.0.1 I have to change constraints with `>=`. Please check the constraints in `package.json` now.

bastianwegge:
Ah okay, solves it! Thanks!

PR #340, opened 2017-09-08
prettier precommit using husky / lint-staged

bastianwegge:
This PR is enabling prettier to reformat all committed code files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "precommit": "lint-staged"
   },
   "lint-staged": {
-    "*.{js,json,css}": ["prettier --write", "git add"]
+    "*.{js,jsx,json,css}": ["prettier --write", "git add"]
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -12,8 +12,10 @@
     "build": "cross-env NODE_ENV=production spin build",
     "exp": "spin exp",
     "start": "node --harmony build/server",
-    "tests": "cross-env NODE_ENV=test PORT=7070 spin test --full-trace --exit \"src/**/*.spec.js\"",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -12,16 +12,22 @@
...
     "migrate": "knex migrate:latest",
     "rollback": "knex migrate:rollback",
     "watch": "spin watch",
     "heroku-postbuild": "rm -rf build && npm run build && npm run seed",
-    "cli": "node tools/cli"
+    "cli": "node tools/cli",
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "*.{js,json,css}": ["prettier --write", "git add"]
```
mairh: shouldn't you add `jsx` extension too?

vlasenko: And `scss` extension

mairh: since we officially support `less`. That should be added too by default

bastianwegge: There currently is no .less file and I don't see any "we officially support less" sign. You're sure?

mairh:
Are you sure you have run prettier on all files? I see only user module from the client side.

package.json
mairh: Well, maybe it's not needed, but we have less support by default using `spinjs`

bastianwegge:
@mairh I'd vote against reformatting the whole codebase on one swoop. At least that was not my intention. I installed a pre-commit hook (as the name of the PR implies) that takes all committed files (of your current commit) and reformats them.

mairh:
That's perfectly fine.

vlasenko:
@bastianwegge Then can we not reformat the files in this PR at all?
I would vote to reformat all the files in one commit right after we merge this PR

bastianwegge:
closing because of master merge that caused all merged files to be reformatted

PR #341, opened 2017-09-08
prettier support for new files

bastianwegge:
added prettier, husky, lint-staged to dev-dependencies and enabled precommit with prettier + .prettierrc

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "printWidth": 80,
+  "parser": "babylon",
+  "singleQuote": true
+}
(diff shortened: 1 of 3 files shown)


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 80,
+  "parser": "babylon",
+  "singleQuote": true
```
mairh: Why don't we go with the default settings? I don't have any strong opinion on using single vs double quotes, but I definitely prefer using the default setting from a third party lib as long as it is safe.

bastianwegge: printWidth + parser are default and just added to show the settings, doubleQuote everywhere is the default, and as far as I can remember single quote is airbnb (so our eslint) default

vlasenko:
Should I merge it, everyone agree?

mairh:
Yes

vlasenko:
Okay, lets rock

PR #361, opened 2017-09-18
Add Flow typechecking

vlasenko:


vlasenko:
@mitjade @mairh Should we merge this into master? Theoretically if users don't want to use Flow typechecking they can easily write JavaScript code without flow types for new files and if they want they can even disable checking for existing files during development by removing `tcomb` from `.spinrc.json` stack and flowtype from `.eslintrc`. The cons for such users is that the code base of the kit will still mentions flow types, which not looks as terrible bad thing for me. Also `flow-typed` dir is a bit of hassle, since it can grow large, but it shouldn't change often (only when flow-typed package is being installed that don't have flow types bundled with it).

mitjade:
@vlasenko I'm for including it. We should add some docs, also how to disable and remove it if needed. I always wrather delete stuff out, than spend days figuring out how set it up.

Awsame work as always!

mairh:
How easy would it be to upgrade the existing forks? What are the steps?

vlasenko:
@mairh It will be very easy. See the first commit in this PR, it has both config changes needed and one example of using Flow Typechecking

mitjade:
@vlasenko Why do all the `imports` now get double quotes?

mitjade:
Maybe it's my IDE acting up.

vlasenko:
@mitjade Double quotes are because of the prettier acting

vlasenko:
@mitjade No, actually prettier outputs single quotes

vlasenko:
@mairh @mitjade Waiting for you guys to review this PR and after that I'm going to merge this into master. Vital changes are in the first commit, the rest of commits are cosmetics / adding flow-typed generated files.

mitjade:
@vlasenko `src/cleint/app/main.jsx` still has double quotes in imports?

vlasenko:
@mitjade Damn, all should be fixed now

vlasenko:
@mitjade Stop, how comes it don't commit my changes...

mitjade:
@vlasenko Looks like if only quotes are changed nothing gets committed?

mitjade:
Something similar happened to my commit?

mitjade:
@vlasenko I got a new commit from you, but still double quotes?

vlasenko:
@mitjade They are fixed in the last commit

mitjade:
@vlasenko Now all OK!

vlasenko:
@mitjade I didn't figure out what is the problem. But will take a look next time when I commit something

vlasenko:
@mitjade Well, I think the problem was that we have used `prettier --write` instead of `eslint --fix`, which generated slightly different output and caused troubles

mairh:
I have had the problem with prettier single and double quote too last night. Actually, when the pre-commit lint fix applied it changes it to single quotes and when running yarn lint it changes back to double quotes.

I am not sure if that is the one time things or an issue with my ide

PR #364, opened 2017-09-19
active navbar link

mitjade:
- update bootstrap
- make nav bar link active

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "apollo-server-express": "^1.1.2",
     "bcryptjs": "^2.4.3",
     "body-parser": "^1.18.1",
-    "bootstrap": "4.0.0-alpha.6",
+    "bootstrap": "4.0.0-beta",
     "cors": "^2.8.4",
     "dataloader": "^1.3.0",
     "expo": "^15.1.0",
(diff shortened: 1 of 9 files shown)


mitjade:
@vlasenko Any objections?

vlasenko:
@mitjade Looks good to me, thanks Mitja

mairh:
Please update cli template as well

vlasenko:
Merging this one has added two new yarn warnings, I think these deps should be added:
```
warning "bootstrap@4.0.0-beta" has unmet peer dependency "jquery@>=3.0.0".
warning "bootstrap@4.0.0-beta" has unmet peer dependency "popper.js@^1.11.0"
```

mairh:
@mitjade Is there any way to keep the parent link active when going into sub-routes?

Check the Heroku demo

https://apollo-universal-starter-kit.herokuapp.com/post/20

The posts link is not active anymore

mitjade:
@vlasenko These are optional dependencies that we do not use currently, so I would not add them.

@mairh The problem is that `Posts` is not a sub-route of `Post`, so I do not know how to achieve this?
I would lease it as is for now, if I figure this out will add an example later.

mairh:
Yeah, I am aware of the reason why it is not working at the moment. I was more interested in finding a solution. Let me see if I can think of a way to achieve that.

PR #368, opened 2017-09-20
Graphql file names

mairh:
Since we use action (verb) in the beginning for (get, add, delete) query/mutations, it looked odd to me that we use the module name in the beginning for subscription/update graphql filenames and actions.

Let me know if this change makes sense to you guys.

### DIFF
--- a/src/client/modules/counter/containers/Counter.jsx
+++ b/src/client/modules/counter/containers/Counter.jsx
@@ -8,7 +8,7 @@ import CounterView from '../components/CounterView';
 
 import AMOUNT_QUERY from '../graphql/getCount.graphql';
 import ADD_COUNT_MUTATION from '../graphql/addCount.graphql';
-import COUNT_SUBSCRIPTION from '../graphql/countUpdated.graphql';
+import COUNT_SUBSCRIPTION from '../graphql/updateCount.graphql';
 
 class Counter extends React.Component {
   constructor(props) {
@@ -39,7 +39,7 @@ class Counter extends React.Component {
       variables: {},
       updateQuery: (
         prev,
-        { subscriptionData: { data: { countUpdated: { amount } } } }
+        { subscriptionData: { data: { updateCount: { amount } } } }
       ) => {
         return update(prev, {
           count: {
--- a/src/client/modules/counter/containers/Counter.spec.js
+++ b/src/client/modules/counter/containers/Counter.spec.js
@@ -2,7 +2,7 @@ import chai from 'chai';
 import { step } from 'mocha-steps';
 
 import Renderer from '../../../../client/testHelpers/Renderer';
-import COUNT_SUBSCRIBE from '../graphql/countUpdated.graphql';
+import COUNT_SUBSCRIBE from '../graphql/updateCount.graphql';
 
 chai.should();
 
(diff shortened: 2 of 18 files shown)


vlasenko:
Subscriptions use naming `countUpdated`, because you receive data when the count is updated. It is not an action, like mutation, rather it is data that you receive when something happened.

mitjade:
@mairh I agree with @vlasenko, but it would be nice if we had a common convention for all.
Proposal:
`Module<Get|Add|Delete|Update|Updated>`
`<get|add|delete|update|updated>Module`

mairh:
How about `edit`? We use that in the naming convention

mitjade:
@mairh You are actually doing an update. You edit a from. But to be honest, I was the one that made the mistake and used edit instead of update in post example :)

mairh:
I think `edit` still makes sense when combining with the React component. Usually, I have seen people name their component like `EditPost.js` not `UpdatePost.js`. Also, if we keep`edit` for React component and `update` for graphql it seems a bit off

```
import POST_QUERY from '../graphql/getPost.graphql';
import POSTS_QUERY from '../graphql/getPosts.graphql';
import POST_ADD from '../graphql/addPost.graphql';
import POST_DELETE from '../graphql/deletePost.graphql';
import POST_EDIT from '../graphql/editPost.graphql';
import POST_SUBSCRIPTION from '../graphql/updatePost.graphql';
import POSTS_SUBSCRIPTION from '../graphql/updatePosts.graphql';
```

vs

```
import POST_QUERY from '../graphql/getPost.graphql';
import POSTS_QUERY from '../graphql/getPosts.graphql';
import POST_ADD from '../graphql/addPost.graphql';
import POST_DELETE from '../graphql/deletePost.graphql';
import POST_UPDATE from '../graphql/updatePost.graphql';
import POST_SUBSCRIPTION from '../graphql/postUpdated.graphql';
import POSTS_SUBSCRIPTION from '../graphql/postsUpdated.graphql';
```

Someone might get a bit confused between `updatePost` and `postUpdated` if we use `update` in place of `edit`

mitjade:
@mairh Fine by me. Maybe @vlasenko can comment on this as well. I would then also change `POST_QUERY` to `POST_GET`. It is also confusing here as well to have once `POST_ADD` and then `import addPost`. Maybe we pick one and use it everywhere. I like `PostAdd` slightly better because you get them sorted by `Post` and not by `Add` on the disk. Again, my mistake to begin with and ether is fine in the end.

mairh:
So eventually you would prefer

```
import POST_GET from '../graphql/postGet.graphql';
import POSTS_GET from '../graphql/postsGet.graphql';
import POST_ADD from '../graphql/postAdd.graphql';
import POST_DELETE from '../graphql/postDelete.graphql';
import POST_EDIT from '../graphql/postEdit.graphql';
import POST_SUBSCRIPTION from '../graphql/postUpdated.graphql';
import POSTS_SUBSCRIPTION from '../graphql/postsUpdated.graphql';
```

vlasenko:
I'd like to use best practices for GraphQL naming conventions. Unfortunately googling them didn't bring any meaningful results. So I would rather use examples from Apollo team, for example:
https://github.com/apollographql/GitHunt-React/tree/master/ui/graphql

mitjade:
@mairh `POST_SUBSCRIPTION ` -> `POST_UPDATED`?

mairh:
@vlasenko, In that case, we shall make 2 changes. PascalCase and use the actual action name.

Something like

`AddPost, AddPosts, DeletePost, PostQuery, PostsQuery, PostSubscription, PostsSubscription`

vlasenko:
@mitjade Yes

@mairh Yes

mitjade:
Fine by me. I would just unify import define to the same format, just capitalised.

mairh:
I'll make another PR with the above info and we can check

PR #370, opened 2017-09-20
Prettier printWidth 120

mitjade:
I would like to set `printWidth` to 120.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,7 +15,7 @@
     "react/jsx-key": 0,
     "semi": 2,
     "prettier/prettier": ["error", {
-      "printWidth": 80,
+      "printWidth": 120,
       "singleQuote": true,
       "parser": "babylon"
     }]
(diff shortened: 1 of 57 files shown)


mitjade:
@vlasenko Would you agree?

mairh:
I am ok with it. Build is failing though

mitjade:
@vlasenko In my first commit I did not commit any changes to `package.json` or `yarn.lock`, but build failed anyway. Tried to regenerate `yarn.lock` and push again, no difference. I get no error running `yarn test`?

vlasenko:
@mitjade Yes, but Travis logs clearly show that `yarn.lock` is broken. Hashes don't match

mitjade:
@vlasenko Any suggestions how to solve this?

vlasenko:
@mitjade This is not our fault. This is rather fault of module authors. They have replaced library archive with different contents. This will defeat all libraries that use their module and use `yarn.lock`, since hash has changed.

To switch to use new contents of the module `yarn.lock` need to be regenerated. But this cannot be done without clearing yarn cache first. So the recipe to escape from situation like this is:
1. Remove `yarn.lock`
2. Clear Yarn cache: `yarn cache clean`
3. Regenerate `yarn.lock`

mitjade:
@vlasenko Thanks for clearing this up. Will merge this now.

PR #371, opened 2017-09-20
extract pubsub into separate file

mitjade:
@vlasenko In one of my projects I need to import `pubsub` to a file that gets called before `modules` exist. So for this to work I needed to move `pubsub` into a separate file and then include it back into `schema.js`.

Would this be something we can add to `master` or should I keep changes like this in my separate branch, that I then use to up-merge?


vlasenko:
@mitjade Sure we should merge this into master. This is a good change I think, since it improves general modularity of the kit.

mitjade:
@vlasenko Will merge when if build passes.

PR #372, opened 2017-09-20
Graphql file name

mairh:
- As discussed and agreed previously here is a PR for `.graphql` file names. I tried to follow https://github.com/apollographql/GitHunt-React/tree/master/ui/graphql as closely as possible.
- Changes are as follows:
  - PascalCase naming convention for `.graphql` files
  - `.graphql` files are imported with CAPS, but keeping the name same as the PascalCase extension
  -  Remove query postfix from actual function name inside `.graphql` query file. Similar to what we have over here https://github.com/apollographql/GitHunt-React/blob/master/ui/graphql/FeedQuery.graphql. Thus converting `return postsQuery.edges.map(({ node: { id, title } }) => {}` to `return posts.edges.map(({ node: { id, title } }) => {}`, which make things a bit more clear.
- Since we use `count` keyword to count the number of records in posts module. I have renamed `count` to `counter` everywhere in `counter` module. Previously, we have a mix of `count` and `counter` definition inside `counter` module.
- You need to run `rm -fR dev-db.sqlite3` since we need to populate new `counter` table.


mitjade:
@mairh Excellent! Only thing that bothered me a bit was `REGISTER` and `LOGIN`, `LOGOUT`. In all other cases we ether append or prepend module name. This are the only exceptions. But I guess `Register` and `Login`, `Logout` are special cases by them selves.

PR #376, opened 2017-09-21
general improvements

mitjade:
- settings for token secret and google analytics
- expose webSocket in in createContext

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,7 +1,9 @@
 {
   "app": {
     "debugSQL": false,
-    "apolloLogging": false
+    "apolloLogging": false,
+    "secret": "secret, change for production",
+    "googleAnalytics": "UA-000000-01"
   },
   "expo": {
     "name": "Apollo Universal Starter Kit",
--- a/src/client/app/Main.jsx
+++ b/src/client/app/Main.jsx
@@ -94,7 +94,7 @@ const logPageView = location => {
 };
 
 // Initialize Google Analytics and send events on each location change
-ReactGA.initialize('UA-000000-01'); // Replace your Google tracking code here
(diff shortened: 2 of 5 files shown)


mitjade:
@vlasenko I'm trying to port one of my projects to the latest version of the kit.

Whenever I find I made changes outside of modules, I am typing to modularise this, so future upgrades would be easier.

vlasenko:
@mitjade Sure, sounds good

vlasenko:
Looks good to me

mairh:
Just for my information. What does the secret token actually do in this kit?

mitjade:
@mairh Secret is used to sign tokens on the server. So if someone would want to inject a different token, it would not be valid unless he knows your secret.

mairh:
Why aren't we using it for the prod Heroku demo?

mitjade:
@mairh What do you mean exactly?

mairh:
As far I understand it is used for the user authentication stuff? Or something else?

mitjade:
Yes.

mairh:
How does the secret token is generated? So for example, if I do a prod release I should put some random token there?

mitjade:
Currently you need to do this by your self. That is why on intentionally named it `secret, change for production`.

mitjade:
There are still a lot of improvements to be made to `auth`, to be really useful, but it is a bug task and I do not have the time do really do it all, plus I am not that experienced in this area and am still learning stuff as I go along :)

mairh:
It is already useful at the moment. A good addition I would say! We can keep improving it when using it in real life projects.

mitjade:
@mairh Thx! Thats the plan.

PR #378, opened 2017-09-22
custom auth example

mitjade:
In this PR the following changes are made.

- Change `app.json` structure to accomodate future `user` example changes.
- Introduced new table entry `secret`, so user refresh tokens can be easily invalidated if needed. Before it was tied to `password`, but in case of other `auth` scenarios this is not the best choice.
- Rename migration files to prefix with `00X`. This is done so you can specify the order in witch they get added. In case we want to add `user_id` to `post`, before this would not be possible, since `post` was aded before `user`.
- Added a custom `auth` example. If you need a client certificate `auth` in combination with `https`.

More `auth` examples can be added similarly like Facebook, gmail, GitHub, ... I intend to add them when I will need it for my projects, currently I only need `cert auth`, so this is the first one I added.

I know there are few people that might find this useful, but still :)
If you want to test it out in development without certificates, I setup so serial gets populated under `__DEV__` by default, if you turn it on in `app.json` -> `"certAuth": true`.

@vlasenko Your thoughts on merging this into master or moving it into a separate example branch?


vlasenko:
@mitjade I don't fully understand what secret is doing in database. Secret keys must be safe, they should not appear in database.

vlasenko:
It looks like the `secret` in database is really a `password`. I don't understand the reason for renaming in this case. It is less obvious name then password

mitjade:
@vlasenko This is just a part of the secret. In code I do `const refreshSecret = SECRET + user.secret;`. This comes in handy if you want to invalidate a certain user refreshTokens.

mitjade:
Later if you have different `auth` scenarios like `Facebook` you have no password.

vlasenko:
@mitjade Could you research a bit naming for long term tokens and short term tokens. I think the field in database even in this case should be named `password`

mitjade:
@vlasenko Will do.

vlasenko:
@mitjade If you rename `secret` -> `password` in database, I think this PR is fine for master

mitjade:
Maybe I can remove `secret` and always use `password` for secret and in case of `Facebook` or other `auth`s, I populate this field also if empty?

vlasenko:
@mitjade Ideally you should use secret + salt + password. You should use `secret` so that only server was able to issue refresh tokens. You should use salt additionally to prevent dictionary and rainbow attacks on user password.

mitjade:
@vlasenko When I generate `password`, I use `bcrypt.hash(localAuth.password, 12);`.

`Bcrypt` has salts built into the generated hashes to prevent rainbow table attacks.

mitjade:
@vlasenko Removed the `secret` and agin use `password`. In case of other `auth` examples, we can then populate `password` filed as well, for additional security on `refreshTokens`.

vlasenko:
@mitjade Great, thank you Mitja!

mitjade:
@vlasenko Can this be merged?

vlasenko:
@mitjade Yes, please go ahead

PR #383, opened 2017-09-25
module config for user

mitjade:
Extract settings for modules to `config`. Re-import in `settings.js`.


vlasenko:
@mitjade Yes, looks good to me. Thank you Mitja

PR #392, opened 2017-09-27
throw error if email not found

vinu:


mitjade:
@vinu This erros are meant to accumulate and throw at he end, so if login failed for multiple reasons all can be shown to the user.

There is this a lot to be improved here, so I appreciate you are trying to help out.

vinu:
@mitjade, thanks, 
here if the email not found, the localAuth becomes null, and the later password compare fails as it tries to access password field of null

mitjade:
@vinu You are correct. I am still evaluating how to best approach error handling so might make a few more changes. Will merge this.

PR #396, opened 2017-09-29
nodemailer as module, registration email conformation

mitjade:
Example of adding a mail support to this kit. I would like to use this for user email conformation, but could be used independently.

To test it you need to set your Gmail `email` and `password` and allow your gmail to [Allowing less secure apps to access your account.](https://support.google.com/accounts/answer/6010255?hl=en)


vlasenko:
I don't see the usage of this mailer in the code. I think when you add the usage, you will have to change architecture of the code a bit. I think there should be some mailer interface and empty implementation of this interface, so that mailer can be removed freely.

src/server/modules/mailer/index.js
```diff
@@ -0,0 +1,6 @@
+import transporter from './transporter';
+import Feature from '../connector';
+
+export default new Feature({
+  createContextFunc: () => ({ transporter })
```
vlasenko: Why to name in `transporter` here, instead of `mailer` or something like that?

mitjade:
@vlasenko I see your point:

To send an email you need to:
```
      context.transporter.sendMail({
        to: 'mail@examole.com',
        subject: 'Confirm Email',
        html: `Please click this email to confirm your email`
      });
```

Not sure how to provide the empty implementation and where to put it?

vlasenko:
@mitjade But you will need to have something like
```
if (context.transporter) {
    ...
}
```
to allow mailer module removed without consequences. For now it's fine, I think.

mitjade:
@vlasenko I will rename `transporter` to `mailer` and add an example of ussage in the `user` module. Would that be good enough for the beginning.

mairh:
Apart from the CR what else I can test over here? Is there any example you are planning to add?

mairh:
ah just saw your last comment regarding example usage in user module.

mitjade:
I will continue the `email conformation` in this PR and we can take a look in the end if anything can be improved before it is merged.

src/server/modules/mailer/index.js
mitjade: done

vlasenko:
@mitjade Yes, that's all for now, thank you Mitja!

mitjade:
@vlasenko @mairh This should now be working in you set the following:
gmail: `EMAIL_USERNAME ` and `EMAIL_PASSWORD ` in `conf/mailer.json` and `confirm` and `sendConfirmationEmail` to `true` in `conf/user.json`

Any comments what to change before this can be merged?

mitjade:
By default this is not enabled, since you need to set `nodemailer` to work, so I do not want to force this being needed to setup, for the example to work. I set `is_active` flag, so I can reuse the same field to if I want to manually disable/enable user account.

mitjade:
@mairh Any comments or can this be merged?

vlasenko:
@mitjade This can be merged

PR #404, opened 2017-10-03
facebook auth

mitjade:
Work in progress...


mitjade:
This should now be working.

To enable this feature, you need to setup an app under https://developers.facebook.com, to get `clientID` and `clientSecret`.
For development you can add a wesite link to `http://localhost:3000/`.

Then you need to set this fields in `config/user.json` under `auth/facebook`. Also check `enabled` to true.

mitjade:
If you previously logged in with `password`, accounts get joined together by `email` if you use the same one for both. Also if you first login with `facebook` and later register by `password`.

mitjade:
I renamed db `local_auth` and `cert_auth` to `auth_local` and `auth_certificate` and added `auth_facebook`. It is really hard to find this tables in db if they do not start with the same prefix.

mitjade:
Would merge this after https://github.com/sysgears/apollo-universal-starter-kit/pull/413.

mairh:
@mitjade You can now take the latest changes from the master branch and make the necessary changes in `.env` file.

mitjade:
Added my own `clientID` and `clientSecret`. This will only work for `localhost`, but I think I can leave this in anyone can test this out out of the box.

src/server/server.js
```diff
@@ -80,10 +80,12 @@ for (const middlewareUse of modules.middlewaresUse) {
...
-  if (middlewareGet.callback2) {
-    app.get(middlewareGet.path, middlewareGet.callback, middlewareGet.callback2);
-  } else {
-    app.get(middlewareGet.path, middlewareGet.callback);
+  if (!isEmpty(middlewareGet)) {
+    if (middlewareGet.callback2) {
+      app.get(middlewareGet.path, middlewareGet.callback, middlewareGet.callback2);
+    } else {
+      app.get(middlewareGet.path, middlewareGet.callback);
+    }
```
mairh: This block is a bit hard to understand. Probably the naming convention is not very clear.

mitjade: Will refactor this part.

mairh:
Maybe add FB on register page as well.

`Register with Facebook`

http://localhost:3000/register

screen shot 2017-10-06 at 15 26 15

mitjade:
@mairh  Why? You would only want to use Facebook login if you are already registered there.

mairh:
hmm just considering UX. If you are a first time user and go to register/signup we are showing only one option. I was considering the medium flow (https://medium.com/) where FB option is available on both signin / signup. However, looking closely, I noticed that they both open the same url. Perhaps, we don't need to add Fb option on register route

mairh:
Do I need to change something?

screen shot 2017-10-06 at 15 36 14

mitjade:
Looks like `clientID` and `clientSecret` will only work for me, while in development mode.

You would need to create your own to test it out.

mairh:
ok then. Can you remove your own credentials from `.env` and replace it with a placeholder

mitjade:
done

mitjade:
@mairh Can you copy paste full medium link?

mitjade:
I agree it does not add much value. But does not do harm as well, so I can add it id this is a standard practise.

mairh:
But I can see they have a bit different login/register flow than us. Since they expect user to use social media login, if not, they check if the email is present then prompt with the password to log in otherwise show the register section. Let's not dig deeper into that.

mitjade:
@mairh Looks like the link gets cut off.

mairh:
Do you wanna add a section in readme on how to enable FB auth? Or you prefer to do it along with Main sending feature when you are planning to refactor user module.

mitjade:
I will add it when I'm done with `user` example, since things will probably still change.

mairh:
Yes, that's ok!

mitjade:
If this get's approved I will move on to `users` and `profile` page.

Being able to add/delete users if admin. enable/disable, sort, filter.

Change password on profile and forgot link on login.

I think this would cover all the minimum essential features, for the user module to be useful.

mairh:
@mitjade 

```
/home/travis/build/sysgears/apollo-universal-starter-kit/src/server/server.js
  6:27  error  'isEmpty' is defined but never used  no-unused-vars
```

mitjade:
@mairh Ok to merge, or have any other concerns?

mairh:
The code looks good to me. Let me just quickly test it on local with my own fb client and secret id

mairh:
What should I add over here

screen shot 2017-10-06 at 16 26 43

I 've created an app on FB developer

mairh:
With the following settings, I am getting the same error

screen shot 2017-10-06 at 16 51 45

vlasenko:
@mairh You should add `localhost:3000` as domain, I think, not just `localhost`, since these are considered to be different domains

mitjade:
I have App domains empty and it works?

mairh:
Check these

https://github.com/specious/facebook-cli/issues/1
https://github.com/specious/facebook-cli/issues/17

mitjade:
:)

mairh:
@vlasenko Can you try to test this PR on your local machine?

vlasenko:
@mairh Yes, give me a minute

vlasenko:
@mairh I was able to succeed by doing the following steps:
1. In the left navbar choose `Add product` -> `Facebook Login`, no need to type anything there just skip to step 2
2. Goto this Product settings.
3. Enter `http://localhost:3000/auth/facebook/callback` under `Valid OAuth redirect URIs`

mairh:
ok, if it works for you then @mitjade can merge this PR. I'll try it out later.

I just wonder why it worked for @mitjade by setting `http://localhost:3000` in site url, whereas for you in product settings.

vlasenko:
@mairh I'm not sure either. @mitjade could you tell us your steps, how did you setup your Facebook app?

vlasenko:
@mitjade PR is good to merge, though

mitjade:
I created a new app and could not make it work ether with the same settings as before, old one still worked?

Tried what @vlasenko recommended and it works! So looks like you need to add `Valid OAuth redirect URIs` under Product settings for `Facebook Login`.

Will mention this when I write the docs on how to use this.

PR #413, opened 2017-10-05
Secrets

mairh:
- The idea behind this PR is to secure important details, which shouldn't go in the code repo.
- People can easily add `secrets.js` in `.gitignore`. All the password and sensitive info are in one place rather all over config folder files.
- I didn't add the `secrets.js` in `.gitignore` in this PR because we don't have any sensitive data.

P.S -> By default the verification is enabled which can be accessed at
https://ethereal.email/

Username and password are in `.env` file


vlasenko:
I would rather use .env file for secrets

mitjade:
+1 for .env

Otherwise I agree this needs to be extracted from configuration files. Also `googleAnalytics` from `app.json`.

mairh:
Would it directly be available from `process.env.xxxx`?

Can I simply replace

```
export default {
  service: config.mailer.emailService,
  auth: {
    user: config.mailer.emailUser,
    pass: config.mailer.emailPassword
  }
};
```

with

```
export default {
  service: process.env.EMAIL_SERVICE,
  auth: {
    user: process.env.EMAIL_USER,
    pass: process.env.EMAIL_PASSWORD
  }
};
```

and in `.env``

```
EMAIL_SERVICE=Gmail
EMAIL_USER=something@gmai.com
EMAIL_PASSWORD=somePassword
```

vlasenko:
@mairh Yes

mairh:
@vlasenko @mitjade Should be ready to go now from my side

mitjade:
Could you also add `googleAnalytics` from `app.json`. Might be a problem because `app.json` is not `.js`?

mairh:
I've added the tracking io .env file

.gitignore
```diff
@@ -6,3 +6,4 @@ npm-debug.*
 .expo/
 *.sqlite3
 .DS_Store
+jsconfig.json
```
mitjade: Do we still need this ignore?

mairh: This is not related to this PR. But in some IDE's (for e.g. VS Code) if you have enabled `jshint` extension this file is added by default to the root folder. This is just for safe keeping

mitjade: OK

mairh:
@vlasenko I will merge this since it is blocking @mitjade FB auth PR. If you have anything to add pls let me know and I will change that in another PR.

mairh:
@mitjade Looks like the `.env` thing is not working. I only tested with the `secrets.js` but didn't test the application when I switched to `.env`. Not sure why the config can't fetch the values from `.env`

mitjade:
@mairh You might need to add https://github.com/motdotla/dotenv.

mitjade:
This might not work for `GOOGLE_TRACKING_ID`.

mairh:
If we add `dotenv` we need to change webpack config in spinjs to pass

```
node: {
  fs: 'empty'
}
```
We were already using `APP_DIR` from `.env` in Docker files. I wonder do we really need `dotenv`?

mitjade:
@mairh I guess we need to wait for @vlasenko to help out here.

vlasenko:
Let me take a look at this one. Though I didn't understand what should I check, where is the problem?

vlasenko:
Okay, I understood

mairh:
@vlasenko @mitjade This is fixed now via https://github.com/sysgears/apollo-universal-starter-kit/commit/bdfd987ab55686680c741daa8e044a69e1a812e9

vlasenko:
Great! Many thanks @mairh !

PR #416, opened 2017-10-09
support sqlite, mysql and pg out of the box

mitjade:
The idea here is to provide easy transition form `sqlite` to ether `mysql` or `pg`, when you want to use this kit on a real project.

To switch, you need to:
-  set `DB_TYPE ` in `config/secrets.js` form `sqlite` to ether `mysql` or `pg`.
- set `DB_USER `, `DB_PASSWORD ` and `DB_DATABASE ` in `.env`

What should this PR solve:
- no need to override `knexdata.js` or any `databases/seeds` files, witch should make `up-merging` easier
- no need do install new packages like `mysql2` or `pg` and add them to `tools/webpackAppConfig.js`
- have all `env` specific settings in one place -> `.env`
- resolves different strategies needed to `truncate` tables: `common/db` -> `truncateTables`


mitjade:
I have now projects using this kit, that are running on `mysql` or `pg`. This would resolve merge conflicts on a few files and make transition easier when switching to `mysql` or `pg`.

mitjade:
Later we could provide `docker` setup for this two databases.

For example `docker-compose.pg.yml`:
```
version: '3'
services:
  postgres:
    image: postgres
    environment:
      POSTGRES_PASSWORD: password
    volumes:
      - ${APP_DIR}/data:/var/lib/postgresql/data
    ports:
      - 5432:5432
```

.env
```diff
@@ -1,6 +1,11 @@
 # Misc
 APP_DIR=/usr/src/app
 
+# Db
```
vlasenko: @mitjade Could you use `DB` instead of `Db` everywhere?

vlasenko:
@mitjade I think we don't need `config/secrets.js` anymore, as all our secrets are now in `.env`, but this is not responsibility of this PR. Otherwise everything looks good to me, great work, thanks Mitja!

mairh:
The reason for having secret.js is to import `dovenv` in one place rather in each and every file where we are using `process.env.xxxx`. However, I guess it could also work if we just import `dotenv` in server.js?

vlasenko:
@mairh Yes, it should work if we import `dotenv` in `server.js`

.env
mitjade: done

mitjade:
I agree regarding `secrets.js` not being needed. We just need to find a place for `GOOGLE_TRACKING_ID`, since it will not work if in `.env`, maybe just move it back to `app.json`?

vlasenko:
@mitjade How about moving all properties from `app.json` to `config/app.js` ?

mitjade:
We can try. Should I merge this and we can continue settings optimisations in a different PR?

vlasenko:
@mitjade Sure, lets merge this PR first

PR #417, opened 2017-10-09
Env Var Refactor

mairh:
- Shall we go with this?


.spinrc.json
```diff
@@ -11,11 +11,11 @@
     },
     "android": {
       "stack": ["react-native", "android"],
-      "enabled": false
+      "enabled": true
```
vlasenko: This file shouldn't be changed

mairh: I think you commented on the previous version before I pushed the update

vlasenko:
@mairh Looks good now, thank you Ujjwal!

mairh:
@mitjade Anything you wanna add?

mitjade:
@mairh All good!

PR #421, opened 2017-10-11
Apollo upload

mitjade:
Example how to add apollo upload to this starter kit.


mitjade:
@vlasenko I quickly thrown together this example and have some concerns, before we decide to integrate it or not.

The library uses `apollo-fetch-upload` that replaces `apollo-fetch`, so I see not easy way to add this as pluggable module.

Would it be ok to keep it like this, or should we somehow detect if this is available and used instead of `apollo-fetch` and have a fallback option if module is removed?
Not quite sure how to do that, so for starters I just replaced it and added it directly to `src/client/app/Main.jsx`, `src/server/middleware/website.jsx` and `src/server/server.js`.

This does than mean that if you remove `uplaod` module, you still need to manually remove stuff in this files and replace back `apollo-fetch` or leave it in and have some extra stuff tied to this kit, that is not used.

vlasenko:
@mitjade We certainly should make upload feature pluggable, the same as we do for other feature-modules. The way to go for this is to use `customFetch` parameter to `apollo-fetch` I think.

mitjade:
@vlasenko I was able to move code out of `src/server/server.js`, but am not able to figure out how to correctly set `createApolloFetchUpload` inside a modue and pass it to `createApolloFetch`. Maybe if I reimplement `createApolloFetchUpload` and pass `constructOptions ` to `createApolloFetch`, but then updates would need to be made manualy and I am sure there is better way.

So when you have some time please take a look. I will try to figure out if I can abstract `apollo-optics` any further.

vlasenko:
@mitjade okay, let me take a look

vlasenko:
@mitjade I've pushed the commit to allow modules supplying `constructOptions` to `apollo-fetch`. This should solve modularity issues. Please review and merge this PR if you think its ready

mitjade:
@vlasenko Thank you. I really appreciate you willing to help out! I agree we should stick with this approach of adding new functionality in the form of modules, even if it means some extra work, to implementing it the right way.

PR #422, opened 2017-10-11
Apollo optics

mitjade:
Integrate apollo optics.

### DIFF
--- a/.env
+++ b/.env
@@ -18,4 +18,4 @@ EMAIL_USER=nxau5pr4uc2jtb6u@ethereal.email
 EMAIL_PASSWORD=7n8dB4pYrrchh452wf
 
 #Apollo Optics
-OPTICS_API_KEY=service:mitjade-apollo-universal-starter-kit:Fue5tm_h6zND0xbPh4F-8Q
+OPTICS_API_KEY=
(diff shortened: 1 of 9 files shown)


mitjade:
@vlasenko Similar problem as with `apollo-upload` here.

Was not able to add `instrumentSchema` and `opticsContext` inside a module and make it work. If this needs to be a module also, I will need some time to figure out why adding this inside a module does not work.

vlasenko:
@mitjade If you want I can take a look to make both features pluggable

mitjade:
@vlasenko Give me an hour or so, if I can figure out this my self, otherwise, when you would have some time, please check if this can be done.

vlasenko:
@mitjade Sure, just let me know if you would like to hand over this task to me.

mitjade:
@vlasenko I was able to clean up some of the code, but not all.

Moved stuff out of `src/server/server.js` and `src/server/api/schema.js`, but still have things in `src/server/middleware/graphql.js`.

I would like to set both `instrumentSchema` and `opticsContext` inside `src/server/modules/optics/index.js`, but it seems there is an issue what gets called first and if I set them in `module` it does not work, but does in `.../middleware/graphql.js`. I commented out my attempt in `.../modules/optics/index.js`.

So again if you will have some time, please take a look how this can be solved, since I am currently not able to figure out a solution.

vlasenko:
@mitjade Sure, will do, thanks!

vlasenko:
@mitjade I propose to close this PR and write integration with Apollo Engine instead:
http://engine-docs.apollographql.com/upgrade-optics.html

vlasenko:
Closed in favor of #423

mitjade:
Agreed! Did not see the docs on this.

PR #420, opened 2017-10-10
Improve user module

mitjade:
Add ability to search, filter and sort users.

Add, edit, delete are still work in progress.

PR #424, opened 2017-10-12
Reset Password

johnthepink:
I've got a couple things to clean up before it's done, but I've made some progress on a password reset feature. I thought I'd go ahead and open a PR to see if there is any feedback :)

Basically, it's sending the user an email with a link containing a one-use-only JWT. That link takes you back to the app and lets you provide a new password.

Let me know if you have any thoughts. I will get the rest of it cleaned up soon.


mitjade:
@johnthepink Great work! Really appreciate your willing to help out.

I am doing some improvements on `user` module as well (#420). Just so we do not do the same work. This was one of the tasks, but have not jet started, so this will work out great, just some merging needed, but should not be a problem.

Will take a look at this today and give you some feedback.

package.json
```diff
@@ -105,7 +105,7 @@
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "react-router-redux": "^5.0.0-alpha.6",
-    "react-transition-group": "^2.2.1",
+    "react-transition-group": "^1.2.0",
```
mitjade: Any reason why to use an older version of this package?

vlasenko:
Awesome work, thank you John! @mitjade, will you have time to review this PR today?

mitjade:
From my side it looks ok and can be merged as is.

@johnthepink you mentioned you will do some cleanup, so I will not merge it yet. Where did you intend to include this links, currently I do not see them being used anywhere.

johnthepink:
@mitjade yep the links are part of the cleanup. I will fix it up later today and then ping you.

package.json
johnthepink: @mitjade Ah yes I had to downgrade this because 2.0+ conflicts with some of reactstrap. I found it when I added an `Alert` to the form. See: https://github.com/reactstrap/reactstrap/issues/498

johnthepink:
@mitjade added:

- inline errors on reset password page
- link to forgot password page from login page
- redirect to login page after successful password change

It should be ready to merge now unless you have any suggestions.

mitjade:
@johnthepink Looks good to me! Will merge, now so I can resolve conflicts with #420, before I merge my PR as well.

mitjade:
@johnthepink I see you only decode token in `resetPassword`. Any reason why not to use `wt.verify`?

johnthepink:
@mitjade good point. Fixed in #432

PR #432, opened 2017-10-13
Verify Reset Password Token

johnthepink:
Use jwt's verify method instead of decode.

From suggestion in #424


vlasenko:
Great, thank you @johnthepink !

Issue #389, opened 2017-09-26
Adding complete Frontend UI template


mairh:
Since we offer a full stack kit with robust backend how do you guys think about adding a complete web UI template?

We already use `reactstrap` and thus we can integrate http://coreui.io/react/ in the kit making it out-of-the-box complete experience on web. Similarly, what we are thinking about `nativebase` for `react-native`

vlasenko:
The general idea is fine. I'm not sure though that it should be core-ui for web, native-base for react-native or something else.

mitjade:
I would actually like to see even less dependency on any particular UI, since this is the area users will most likely prefer various options. Currently the first thing I need to do is remove `bootstrap`, `reactstrap` and `sass`, since I do not use it in any of my projects.

I managed to port 2 of my projects to latest kit version and this is where I get most conflicts. That is why I have a separate branch where I first resolve this conflicts and then up-merge to my projects.

vlasenko:
@mitjade We need to make kit even more modular and introduce concept of installable modules, when you can just install modules you need into your project from the kit or update them to the latest versions.

mitjade:
@vlasenko Agreed. But the moment we remove examples from the kit and move them to a separate repo, there is a high possibility they will not be as up to date as now, since now you are forced to update everything. I also like to keep them around and just disable them, so I can copy paste code I want to reuse somewhere. Since it is really simple to remove them now this is not really an issue for me.

But making the rest of the kit even more modular and easier to update would be really great. Currently I managed to extract all the changes I had to make to this kit into `modules` so the only area I still get potential conflicts are:
- `package.json`, `webpackAppConfig.js` and `yarn.lock`, because of different packages needed
- `knex`:  `sqllite` to `mysql`
- `UI` : `bootstrap, reactstrap, sass` to `antd, less`
- `user`: I reused the user module and changed it to my requirements
- `.spinrc.json` and `app.json`: different than default settings
- `tools/templates`: `UI` specific changes

But this are now all quite easy to merge, since most are settigns.

Areas we could improve this even more:
- provide `mysql`,  `postgres` along side `sqlite`
- extract all `UI` related design outside `modules` (`src/app`, `common/components`) with `styled-components` and have them in one place to override `UI` stuff
- more general `user` module, since this will most likely be customised and reused across projects

vlasenko:
[...]

PR #441, opened 2017-10-18
Extract UI components into common module

mitjade:
Initial try to extract button component into common module (not working)


mitjade:
@vlasenko I would like to tackle the last major thing that causes a lot of merge conflicts for me.

I would like to extract all UI specific components into `modules/common`, like we have now for `react-native` for `buttons`. This way you would have all UI specific stuff in one module and only have to rewrite this, to switch to a new UI like `material design`, `semantic-ui`, `antd`, etc...

But currently if I include `Button` component in my web components like `import { Button } from '../../common/components';`, it seems to also wants to load `react-native` components. I have to comment out all `react-native` components for this to work.

If I include `Button` like `import Button from '../../common/components/Button';` it all works ok.

Aly idea how to fix this? You can see my attempt of this in `counter` example.

vlasenko:
@mitjade I'm not sure I understand the problem. Web front-end will load `react-native-web` components, and what is the problem with this?

vlasenko:
@mitjade If there some non react-native-web components, their files should clearly state that - they need to have `.native.js` extension if they are `react-native`-only components

mitjade:
@vlasenko Thank you for the hint. I was using `react-naitive` specific code in a component that did not have `.naitive.jsx` suffix.

vlasenko:
@mitjade Yes, sounds good. Lets make antd example to master

mitjade:
@vlasenko This PR is now almost ready. Some changes needed for `antd`, but since `bootstrap` is set by default, this could be done later. Also it is still not yet 100% react 16 compatible.

There are quite a lot of changes needed to be made, but I think it will be worth it in the long run. Now if you want to switch from `bootstrap` to `antd`, all you need to do, is some minor changes in `src/client/modules/index.js`, `src/server/modules/index.js` and `src/client/modules/common/components/web/index.jsx`. I'm sure this could be optimised, but I guess it is good enough for the start.

To switch to something like `semantic-ui`, that is not yet implemented, you need to copy `ui-bootstrap` components and reimplement them in `ui-semantic` module. I think this should take less than one hour and you do not need to change one file in any of the other modules to archive this. Similarly `NativeBase` could be implemented now for `react-native`.

This structure in `common` modules is setup now in a way that you can add `universal`, `native` or `web` components. There is some more work to set this up, but in this way, moving components around and using different `ui` packages, would make this easier to achieve. 

With this, I now addressed all points I made in #389, with the files that caused me merge conflicts.

PR #458, opened 2017-10-23
clean up comments, add key to menu item

mitjade:
@vlasenko I would like to remove this comments, because every time I use `cli` the first thing I do is remove comments, since they do not add any real explanation value.


vlasenko:
@mitjade Generally comments like these should not exist at all - they don't have any positive value, so I'm completely agree - they should be removed

PR #466, opened 2017-10-26
Ui refactor

mitjade:
Remove unneeded backend module.

`Antd` refactor imports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ If you are using Apollo v1 please use [starter kit from `apollo1` branch](https:
 
 > Apollo Universal Starter Kit is a SEO friendly boilerplate for [Universal] Mobile and Web app development 
 > built on top of [Apollo], [GraphQL], [React 16], [React Native], [Expo], [Redux], [Express] with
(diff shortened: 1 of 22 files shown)


mitjade:
@vlasenko Is there a way to dynamically change the import in `src/client/modules/common/components/web/index.jsx`?

mitjade:
I would like to write some docs how to switch between supported `UI's` and it would really be cool to be able to make the `UI` switch with just one change.

mitjade:
But even if this remains as is, this way you could load parts of different `UI's` and use more than one. You would need to only load specific parts of css then as well, so common styling do not clash.

vlasenko:
> @vlasenko Is there a way to dynamically change the import in `src/client/modules/common/components/web/index.jsx`?

Yes, webpack support dynamic require syntax, we can utilise it for that purpose

mitjade:
Will read it to it and try to implement it.

mitjade:
@vlasenko Could you give me an example for to accomplish this. I am not able to make this work.

I need for example to import from `../../../ui-[module]/components/PageLayout` and export it as `export { PageLayout }`.

mitjade:
I will just leave it as is, since it only complicates the whole thing and you really do not gain much.

PR #467, opened 2017-10-26
User roles

mitjade:
Provide basic role permissions with the help of `graphql-auth`.


mitjade:
@vlasenko I added basic role permissions and fixed many bugs that I found a long the way. Any objections or can I merge this now?

PR #480, opened 2017-10-30
[WIP] Domain driven design for users module

vlasenko:
This is an experiment to make domain entities definitions - a single source of truth for users module


mitjade:
@vlasenko I see you are making good progress with this PR. I have a couple of questions regarding current implementation.

I see you are using [`change-case`](https://github.com/blakeembrey/change-case), but I already used [`humps`](https://github.com/domchristie/humps) in `src/server/modules/user/sql.js`. I can rewrite it to use `change-case` if you prefer it.

Now that knex tables are generated through `domain schema`, how would then creating a separate migration file work, if needed later when in production?

Do you intend to add [`simple schema`](https://github.com/aldeed/node-simple-schema) and use it for validation, or are you trying to implement it without.

You created a separate project for [`domain schema`](https://github.com/sysgears/domain-schema). Do you intend to move code then there, when it is finished?

What else do you still plan to implement, before this can be merged. I am eager to start playing around with this and am just curious.

vlasenko:
@mitjade 
Hi Mitja!
Good questions!
  1. I will check `humps` and probably switch to it, thanks.
  2. As for migrations I think `domain schema` just need to print the code for migration and you need to copy paste it to migrations file for now. Making `domain schema` fully migrations aware when you change schema definition would be difficult, but it is something to consider for the future, not really a priority at this stage I think. Still code generation for migration from schema definition is nice to have
  3. I'm not going to include simple schema, rather going to reimplement it, because I am using it in one of my projects already and I see its weak points. It is really a bit too tied to Mongo and its a bit outdated. I would like to make simple-schema refresh. The core idea is to utilise ES6 syntax to make schema definitions more readable. I am using ES6 property initializers already for this purposes and Array fields look much better than in original `simple-schema`
  4. Yes, I intend to move the code into separate `domain schema` package sooner rather then later.
  5. I am going to implement validations, mostly reimplement ideas from SimpleSchema, but to a degree. I think they are overcomplicated in SimpleSchema. And I would like to have forms generation from `domain schema`

mitjade:
@vlasenko Great! Let me know if I can help in any way.

vlasenko:
@mitjade Okay, let me prepare grounds so that we can work together to push this feature faster.

src/server/modules/user/resolvers.js
```diff
@@ -27,6 +27,14 @@ export default pubsub => ({
       }
     }
   },
+  User: {
```
vlasenko: @mitjade Mitja, I wonder why this is needed? The resolvers seems to be defined as identity functions - they return their argument. I thought if they are absent - everything should work the same? So what is the catch?

mitjade: @vlasenko Not sure, but if I remove this `UserProfile` is not called.

PR #481, opened 2017-10-31
User Subscriptions

johnthepink:
The purpose of this is to add a user subscription module to the kit so that users can subscribe to paid content. This is still a work in progress, but I wanted to get a PR going to get feedback from everyone before I get too deep in. Let me know your thoughts on this. :)

Current progress:

- [x] setting in config to turn on/off subscriptions
- [x] redirect to subscription form after register if setting enabled
- [x] migration to add subscription table with user foreign key
- [x] `SubscriberRoute` which wraps `AuthRoute`
- [x] `SubscriberNav` which wraps `AuthNav`
- [x] `subscribe` mutation which don't actually process anything yet, but does create subscription for current user
- [x] subscription form UI
- [x] private route `/subscribers-only` which requires a subscription
- [x] move subscription to separate migration for modularity
- [x] example subscriber only query
- [x] add card expiration date to form and mutation
- [x] add payment processing using Stripe
- [x] cancel
- [x] webhook for subscription inactive
- [x] generate email for canceled subscription
- [x] generate email for failed charge
- [x] update profile for subscription
- [x] update card info (after update, after  #cancel)
- [x] update styles since branch
- [x] cleanup

There is a subscriber account created:

email: subscriber@example.com
password: subscriber


mitjade:
@johnthepink Great work!

I would move `subscription` table to a separate migration file, so it would be easier to remove this module, if not needed.

vlasenko:
@johnthepink Looks great so far! Awesome work!

johnthepink:
Thanks! I'll keep this rolling then.

@mitjade I like your suggestion, and I'll make that change.

mairh:
Would it be possible to enhance the feature to either subscribe for paid content or one-time transaction/payment?

I could think of an example in e-commerce or any other product related apps where you can hook up the payment processing service credit card/stripe and do one-time payment along with the subscription model.

johnthepink:
@mairh I don't see why not. I'm focused on the subscription aspect right now, but then I could make another pass and extend the module if everyone thinks that's a good idea. I'd be happy to do it.

johnthepink:
I haven't forgotten about this. Been out of pocket. Will get back to finishing this up later this week or early next week.

vlasenko:
@johnthepink Cool, thank you John!

johnthepink:
I think this is ready for review. Let me know what thoughts you have.

I feel like I need to document some of this, but wasn't sure where the best place was. What do you think?

To test it out:

1. get a Stripe account
2. add your publishable key and secret key to the settings
3. run `yarn stripe:setup` to create the subscription plan
4. start the app

There is a subscriber test account, but it is dummy data. Maybe I should give that account a real subscription in the setup script.

mitjade:
@johnthepink Great work! Will try to take a look today and give you some feedback.

vlasenko:
@johnthepink Much of hard work, thanks John! I think you can put docs into `src/client/modules/subscription/README.md` and we can link to them from root `README.md` for now.

johnthepink:
@mitjade @vlasenko thanks! I will get something written up, and also rebase these conflicts that have shown up.

johnthepink:
README added and rebased!

mitjade:
@johnthepink Sorry for a late reply. You really done a tone of work here! Will leave some comments in the code.

mitjade:
@johnthepink There are quite a few places this module adds code outside it's module structure. We are trying to make every module to be as independent form the rest of the kit as possible. In case you wish to remove the module, ideally there should be no remaining code left to clean up. I guess we will have to require certain modules as dependencies, like this one does `user` module.

@vlasenko What are your thoughts on this. Some things can be quite hard to abstract. Can certain exceptions be made, so more complex examples like this one, can be added or should we try and enforce this, so we can later more easily make this as plugins and make them installable through `cli`?

src/server/middleware/html.jsx
```diff
@@ -33,6 +34,7 @@ const Html = ({ content, state, assetMap, css, helmet, token, refreshToken }) =>
           />
         )}
         {!!css && css}
+        {settings.subscription.enabled && <script src="https://js.stripe.com/v3/" />}
```
mitjade: Try and move this to module and abstract it similarly as `stylesInserts` is doing for styles.

src/server/index.js
```diff
@@ -21,3 +22,11 @@ if (module.hot) {
 
   module.hot.accept();
 }
+
+if (__DEV__ && settings.subscription.enabled) {
```
mitjade: Try moving this to module and abstract it similarly as `middlewares` is ding for `app.use`.

src/client/modules/user/components/LoginView.web.jsx
```diff
@@ -48,6 +48,7 @@ class LoginView extends React.PureComponent {
               <CardTitle>Available logins:</CardTitle>
               <CardText>admin@example.com:admin</CardText>
               <CardText>user@example.com:user</CardText>
+              <CardText>subscriber@example.com:subscriber</CardText>
```
mitjade: Add a `settings.subscription.enabled` here as well.

vlasenko:
@mitjade Yes, would be great if you mark all such places and I can help with adding new hooks into module system. I think we should try to make all features modular, so that we can install/uninstall them via cli later.

mitjade:
@vlasenko Should we then also add such hooks in for example `user` module, so code can be inserted from other modules?

package.json
```diff
@@ -23,7 +23,8 @@
     "heroku-postbuild": "rm -rf build && spin exp login -u $EXP_USERNAME -p $EXP_PASSWORD --non-interactive && spin -c .spinrc.prod.json build && spin exp p --non-interactive && npm run seed",
     "precommit": "lint-staged",
     "cli": "node tools/cli",
-    "postinstall": "opencollective postinstall"
+    "postinstall": "opencollective postinstall",
+    "stripe:setup": "node tools/stripeSetup"
```
mitjade: For things like this, we might need an uninstall script, to clean this up, after you do a `yarn cli deletemodule XXX`?

src/client/modules/user/containers/Register.jsx
```diff
@@ -28,6 +30,9 @@ const RegisterWithApollo = compose(
             return { errors: register.errors };
           }
           if (history) {
+            if (settings.subscription.enabled) {
```
mitjade: Maybe `user` module needs hooks, so this can be overridden in specific module?

tools/stripeSetup.js
```diff
@@ -0,0 +1,22 @@
+require('dotenv/config');
```
mitjade: This should be moved to server module.

vlasenko:
@mitjade Yes, we should

tools/stripeSetup.js
vlasenko: @mitjade Are you sure? I think `tools/stripeSetup.js` is used from `package.json` script `stripe:setup`

mitjade: @vlasenko Can't we change the path there also?

vlasenko: @mitjade Sure we can, but should we put this code into server module, is it right? This is technically not a part of backend, but rather a helper script

mitjade: @vlasenko I guess we need to define a way how to handle this kind of cases, since it is something other complex modules will likely require as well.

src/client/modules/user/components/LoginView.web.jsx
johnthepink: Added

tools/stripeSetup.js
johnthepink: My thoughts were the same. I wasn't sure where it needed to live, `tools` seemed better than client or server. Although, that does kind of break the modular concept. I'm happy to move it wherever you decide.

src/server/middleware/html.jsx
johnthepink: Great idea. Added.

src/server/index.js
johnthepink: Agreed, and added.

johnthepink:
@mitjade @vlasenko thank you for taking the time to look this over! Sorry, I know it is a lot of code.

I think I've taken care of the issues I can at this point. Let me know if I missed anything.

@vlasenko are you saying you are going to get a module hook system going that I can tie that registration redirect in to? Let me know if you need any help.

As far as the stripe setup script goes, if it can't live cleanly anywhere at this point I'd be happy to remove it and supplement the documentation. It's very easy to do manually so I don't think the automation is all that useful unless we add some more complicated scenarios.

mitjade:
@johnthepink Great work. I think this is good enough for now. We can come back later and do more abstraction if needed.

vlasenko:
@johnthepink Many thanks John for this work. This is landed into master now. I have granted write permissions to this repository, so feel free to push small fixes directly into master branch or create branches inside this repo when you want to work on a new PR

johnthepink:
@mitjade @vlasenko awesome. Thanks so much!

PR #497, opened 2017-11-09
Send email on add new user with login details

mairh:
Send an email with login details to the user when added or registered.

### DIFF
--- a/config/user.js
+++ b/config/user.js
@@ -6,6 +6,7 @@ export default {
     password: {
       confirm: true,
       sendConfirmationEmail: true,
+      sendAddNewUserEmail: true,
       enabled: true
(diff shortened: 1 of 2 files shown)


mitjade:
@mairh Looks great! Maybe make a function and use it in both places.

mairh:
I think instead of a function we can create an email-template folder and have all the emails in one place. But that will require some refactoring. I guess we can go with the current solution and creating a function wouldn't make things easier for users who would like to change some minor things in the current template.

vlasenko:
Thank you @mairh, looks good!

mitjade:
@mairh Yes, we can deal with this later on.

PR #509, opened 2017-11-15
Replace reactstrap with common components

mitjade:


mitjade:
@vlasenko I replaced `profile` `table` with `card`. It does not look as good, but I do not jet have a solution how to do this in `antd`.

vlasenko:
@mitjade It looks fine with the `card` too, thanks!

PR #522, opened 2017-11-22
[WIP] Implement authentication across all platforms

vlasenko:
Add authentication across all the platforms into `user` module.


theWeird:
Does login already work?
I am looking forward to it for a current project at university ;)

Do you have a reason for removing the code submitting the JWT generated tokens?

vlasenko:
@theWeird Well, I am going to refactor and add JWT code back a bit later in this PR, just wanted to do cross-platform auth based on sessions, because implementation will be more simpler and it is easy to make sessions more reliable, so I want to rather turn on sessions as the default implementation for the kit, but have a JWT-based auth as another module too. I think I will be unable to finish this PR quickly, so feel free to start utilizing parts from it in your project if you find them useful

theWeird:
Do you have an idea on how to handle different navigation options and shown screen components based on login status for mobile?

I did some experiments with IsLoggedIn and IsNotLoggedIn for mobile navigation items.
e.g. show different screens (login/logout screen) or don't even show a item if not logged in.  
It seems to me, that for creation of the navigator objects I am bound to Json objects and plain JS logic and therefore cannot use wrapper components.

For sending requests to Apollo, I also tried to extract the appollo client creation from Main.jsx to a seperate module, but it still relies on being loaded in context of a react module. So I wasn't able to import it in my client/module/user/index.jsx. 

Do you have an idea how this can be handeled?

vlasenko:
@theWeird 
1. To handle navigation screens based on login status this approach should be used:
https://medium.com/the-react-native-log/building-an-authentication-flow-with-react-navigation-fb5de2203b5c
2. When you have extracted Apollo Client creation from Main.jsx, don't forget to handle server side rendering part too in `src/server/middleware/website.jsx`

vlasenko:
Closed in favor of #644

PR #524, opened 2017-11-23
[WIP] Cli crud

mitjade:
Create a cli to generate full CRUD out of a given schema.

`yarn cli addmodule modulename` - to generate an emply module, with no graphql
`yarn cli addcrud modulename` - to generate the crud module
`yarn cli updateschema modulename` - to re-generate schema definitions
`yarn cli deletemodule modulename` - to delete a module


verdverm:
This is neat, have you seen my project https:github.com/hofstadter-io/geb ?  It's a dsl-to-anything generation tool. lots of flexibility and conditionals at the project, file, and module levels. Still need to add the `.hofignore` was going to gebbify this project at some point ;]

packages/client/src/modules/common/components/crud/FilterView.web.jsx
```diff
@@ -4,7 +4,7 @@ import { Formik } from 'formik';
 import { DebounceInput } from 'react-debounce-input';
 
 import { hasRole } from '../../../user/containers/Auth';
-import { Form, FormItem, Input, Row, Col, Button, Icon } from '../web';
+import { Form, FormItem, Input, Row, Col, Button, Icon } from '../web/index-antd';
```
vlasenko: Why `index-antd.jsx` is needed at all? Why not importing `../web/ui-antd` without adding dumb file? The second question - is why are we depending on `ui-antd` here?

Werter12: I'll change import, dump file not needed. We depending on ui-antd because all CRUD cli templates needed it, in contrast whole app use bootsrap-ui and imports from/web/index.jsx hence we can't import index both for CRUD and app components because conflicts of names among those two libraries stop us.

vlasenko: @Werter12 No, this is not true. The app is importing all the components from `web/index.jsx`, but `web/index.jsx` is exporting either AntD or Bootstrap components. To change this you just uncomment/comment one line in index.jsx. Your changes has broken this mechanism. Please revert your changes

vlasenko: @Werter12 We cannot merge this PR until it will support all the UI's we have in the kit: AntD, Bootstrap, AntD mobile, Native Base. Please work on adding this support

mitjade: I think some features are a bit `antd` specific, so I think we could add support from other UI's in a more limited scope, meaning not having all features available. Inline editing and sorting are an example of that functionality.

vlasenko: @mitjade Yes, thats fine I think. But generated code should generally work for all UIs, though with limited functionality on some UIs

mitjade: @vlasenko Yes. I agree.

tools/cli/module.js
```diff
@@ -409,26 +409,32 @@ input ${pascalize(value.type[0].name)}UpdateWhereInput {
...
+      const replaceFragment = `${Module} {(.|\n)*\n}`;
+      fragmentGraphql = shell
+        .cat(fragmentGraphqlFile)
+        .replace(RegExp(replaceFragment, 'g'), `${Module} {\n${fragmentGraphql}}`);
+      try {
+        fs.writeFileSync(pathFragment + fragmentGraphqlFile, fragmentGraphql);
+      } catch (err) {
+        logger.error(chalk.red(`✘ Failed to write a ${pathFragment}${fragmentGraphqlFile} file!`));
+      }
+      logger.info(chalk.green(`✔ Fragment in ${pathFragment}${fragmentGraphqlFile} successfully updated!`));
```
vlasenko: This message will be printed, even if exception occured

tools/cli/helpers/util.js
```diff
@@ -58,6 +58,17 @@ function computeModulesPath(location, moduleName = '') {
   return `${BASE_PATH}/packages/${location}/src/modules/${moduleName}`;
 }
 
+/**
+ * Run prettier on file that was changed.
+ *
+ * @param pathToFile
+ */
+function runPrittier(pathToFile) {
```
vlasenko: runPrittier -> runPrettier?

mitjade: Will fix

PR #530, opened 2017-11-28
Add native-base / antd-mobile for native UI

mitjade:
- refactored post list to use FlatList
- implemented infinite scroll with onEndReached
- added swipe action for deleting post
- refactored edit post to use native-base / antd-mobile components


src/client/modules/common/components/native/ui-native-base/components/FormButton.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyleSheet, Text } from 'react-native';
+import { Button } from 'native-base';
+
+const FormButton = ({ children, ...props }) => {
```
mairh: I guess this should be only `Button`. It is not specifically tied to `Form` component.

mairh: ah I see you have another component named `Button`

mitjade: @mairh I needed to style the button a bit differently if in form, so I added another wrapper.

mitjade:
@mairh @vlasenko Are you OK with merging this to master?

Issue #496, opened 2017-11-09
Secure /login and /register route for logged in user


mairh:
At the moment a logged in user can access `/login` and `/register` route by entering it manually in the url address bar, which should not happen.

bearcanrun:
Made a PR attempt for this at #543. Let me know what you think!

PR #543, opened 2017-12-03
Secure /login and /register route for logged in user

bearcanrun:
Thanks for the awesome boilerplate! In response to #496:
1.  Created an `AuthLoggedInRoute`: Protects any desired routes when the user is already logged in such as _/login_ or _/register_ and redirects to desired location (example: _/profile_).
2. Created an `AuthLoggedIn` navlink: Abstracted version of `AuthProfile`.  Uses minimum `to` and `label` props and I should probably make those required. Would be good to have documentation of it somewhere as I didn't create a sample link and would be happy to if you have a recommendation where. Example: `<AuthLoggedIn to="/profile" label="Profile" className="nav-link" activeClassName="active" />` 

If there's any preferred naming, just let me know and I'd be happy to replace. Cheers!

### DIFF
--- a/src/client/modules/user/containers/Auth.jsx
+++ b/src/client/modules/user/containers/Auth.jsx
@@ -79,6 +79,15 @@ AuthNav.propTypes = {
   cookies: PropTypes.instanceOf(Cookies)
 };
 
+const AuthLoggedIn = withCookies(({ cookies, ...rest }) => {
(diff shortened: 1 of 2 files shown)


vlasenko:
Looks good to me, thanks!

@mitjade Does it looks good to you?

mitjade:
@vlasenko Looks ok from my side.

PR #546, opened 2017-12-05
Upload files improvements

mitjade:
- upload multiple files
- list uploaded files
- remove files
- download link


mairh:
Nice. I was thinking maybe write a real life example. We could upload a profile image and save the url link with meta data to user profile?

mitjade:
@mairh I would rather keep examples separate, since then you are tying `upload` and `user` example together.

mairh:
Yeah I was wondering about the same point. Maybe we keep it separate

package.json
```diff
@@ -78,6 +78,7 @@
     "error-stack-parser": "^2.0.1",
     "expo": "^23.0.0",
     "express": "^4.16.2",
+    "filesize": "^3.5.11",
```
mairh: This one is not added to `webpackAppConfig.js`

mitjade: It is really small and can be used on all platforms, so I did not add it.

PR #553, opened 2017-12-07
enable google oauth

verdverm:


src/server/modules/user/index.js
```diff
@@ -168,5 +219,52 @@ export default new Feature({
...
+
+    // Setup Google OAuth
+    if (settings.user.auth.google.enabled) {
+      app.use(passport.initialize());
+
+      app.get(
+        '/auth/google',
+        passport.authenticate('google', {
+          scope: [
+            'https://www.googleapis.com/auth/userinfo.email',
```
vlasenko: @verdverm Maybe its better to move these settings into `config/user.js` -> auth -> google ?

verdverm: Done, did the same for the Facebook items.

src/server/modules/user/index.js
```diff
@@ -65,6 +66,60 @@ if (settings.user.auth.facebook.enabled) {
...
+      },
+      async function(accessToken, refreshToken, profile, cb) {
+        const { id, username, displayName, emails: [{ value }] } = profile;
+        try {
+          let user = await User.getUserByGoogleIdOrEmail(id, value);
+
+          if (!user) {
+            const isActive = true;
+            const [createdUserId] = await User.register({
+              username: username ? username : value,
```
verdverm: I used the email here because I hit the unique constraint on username in the database.

I was logging in with two of my gmail accounts where my display name (`fname + lname`) is the same, but my email is different.

src/server/modules/user/index.js
```diff
@@ -65,6 +66,60 @@ if (settings.user.auth.facebook.enabled) {
...
+              userId: createdUserId
+            });
+
+            await User.editUserProfile({
+              id: createdUserId,
+              profile: {
+                firstName: profile.name.givenName,
+                lastName: profile.name.familyName
+              }
+            });
```
verdverm: Setting the local user profile from the google profile information.

Should we do the same for Facebook?

src/server/modules/user/index.js
```diff
@@ -65,6 +66,60 @@ if (settings.user.auth.facebook.enabled) {
...
+              isActive
+            });
+
+            await User.createGoogleOuth({
+              id,
+              displayName,
+              userId: createdUserId
+            });
+
+            await User.editUserProfile({
```
verdverm: Setting the local user profile from the google profile information.

Should we do the same for Facebook?

verdverm:
Generally, I think there is a more generic / parameterized method for enable all the passport strategies. That can be left for another PR.

src/server/modules/user/index.js
vlasenko: @verdverm Yeah, I'm not sure why do we need `username` at all. I think `email` should be enough, for email/password based authentication. @mitjade Mitja maybe you have idea why unique `username` is needed now in `user` module?

src/server/modules/user/index.js
vlasenko: @verdverm Yes, would be good, if we can access this data freely from Facebook without asking permission from user.

vlasenko:
Agreed, we should unify different authentication strategies as much as we can.

src/server/modules/user/index.js
verdverm: I would say `email` is for identity and `username` is more of a display name within an application. Users should generally be able to change their display name, and the account is tied to the email.

Facebook requires your username to be your real name, however Reddit does not. It is also possible to tie multiple emails / login providers together into one account in an application by email(s).

I say we leave it for now and deal with it while working on the `rbac` or auth overhaul.

src/server/modules/user/index.js
verdverm: The scopes (per oauth provider) determine what the user sees and agrees to give access to.

There are probably two sensible defaults in general, `email only` and `email + basic profile`.  Then people who use this project can add configuration to expand the scopes they need. We could provide a hook in the auth cycle to extract and store extra info during registration too. Many oauth permission grants are used with apis and I would expect this is more application dependent.

vlasenko:
@verdverm  Thank you for this work. This is landed into master now. I have granted you write permissions to this repository, so feel free to push small fixes directly into master branch or create branches inside this repo when you want to work on a new PR

src/server/modules/user/sql.js
```diff
@@ -18,12 +18,15 @@ export default class User {
         'up.last_name',
         'ca.serial',
         'fa.fb_id',
-        'fa.display_name'
+        'fa.display_name',
+        'ga.google_id',
+        'ga.display_name'
```
mairh: I think this will override the facebook `display_name` since both tables are returning the same column name.

vlasenko: Yeah, I think you are right. Any thoughts, @verdverm  ?

mairh: We need to give them aliases

mairh: This will probably fix it https://github.com/sysgears/apollo-universal-starter-kit/commit/3bc8293e0d0470676b69ec8e57aa7dc00d7a7f47

PR #557, opened 2017-12-08
[WIP] Chat module

verdverm:
Demonstrates a `chat` module which incorporates the `user` module and real-time communication from subscriptions.

---

- [x] prototype implementation
- [ ] graphql permissions / scopes
- [ ] database seeds
- [ ] spec files and tests
- [ ] documentation


vlasenko:
@verdverm Could we use Gifted Chat in this PR?
https://github.com/FaridSafi/react-native-gifted-chat
It looks having good UI and features. I know it is available only on Mobile platforms, but not on Web. But since we have React Native Web integration we can port Gifted Chat to Web too. It is quite an effort of course, but result will pay off in the long run, I think

verdverm:
I'm all about not reinventing wheels. I'd prefer to see the project pull in many third-party components and make them accessible through configuration.

I decided to use `react-table` for table views in this vane. Thought, at the time I made this PR, I was thinking some of the modules in the starter kit are more examples of how to do things than they are like wordpress plugins.

At some point we will probably want to break out the examples, modules, and plugins, into some library or repository kind of thing.

vlasenko:
Closing this PR, because work seems frozen on this one

PR #562, opened 2017-12-09
[WIP] - Dr. Worm's Madd Science

verdverm:
Experiments which add, enhance, modularize, and make configurables all the things

- [ beta ] users, groups, orgs, service accounts
- [ beta ] auth(n/z) methods.
- [ alpha ] RBAC &  fine-grain controls over who can do what in the application.
- [ prep ] Multiple storage engines

---

You should be able to use the same getting started commands to test out this branch.


mairh:
Just wondering what is the motivation/idea behind the entities level auth? I can completely see the benefit of having role based auth, but having entities is a very application specific thing, not a template feature I guess. I could be wrong though.

verdverm:
@mairh Thank you for the question. This will facilitate discussion and design.

I think almost every major application eventually adopts this pattern. The intention is to make each entity (dis/en)able-able. This way we provide the broadest scope which can easily be scaled down.

The motivation is that

- `Orgs` represent Business entities or other legal groupings of individuals
- `Users` represent humans one-to-one
- `ServiceAccounts` represent machines that act on our behalf
- `Groups` are the crossroad of membership and permissioning and act as a "functional group" within the system.

`Orgs` and `Users` are the entities tied to money, whether paying or getting paid. This is application dependent for sure.

`Groups` tie `Users` and `SvcAccts` to permissions on objects or resources. In practice, you don't want permissions on individual entities. You want them on groups and then give individual entities access by `Group` membership. `Groups` are given one or more `Roles` for a set of objects. `Roles` have permissions for making actions against a type of resource. Where ownership of (resources / objects / permissions on) ends up is still a little fuzzy.  

I'm using the Kubernetes RBAC: https://kubernetes.io/docs/admin/authorization/rbac/ as one of my guides.

Thoughts?

verdverm:
we will see! I am hackathoning on this project for the rest of the year ;]

vlasenko:
@verdverm Sounds great, thank you!

verdverm:
Longer-term, the RBAC mechanisms should be pluggable, as most things. (I draw a lot of inspiration for Kubernetes design these days)

RBAC should really be it's own server / database. Then you could plug in LDAP or ActiveDirectory, or use the permissioning systems from Google or GitHub. It is also better practice to separate all the auth stuff into it's own server and have all your applications hook into that. Good for security and cross-application account management.

verdverm:
Two interesting Authorization projects:

- https://github.com/casbin/casbin
- https://github.com/yahoo/athenz

mitjade:
@verdverm In the upcoming changes to this kit, we will be switching from `dataloader` to `graphql-resolve-batch` (https://github.com/sysgears/apollo-universal-starter-kit/pull/580).
Just wanted to give you a heads up, since I noticed you used `dataloader` quite extensively in your branch.

verdverm:
@mitjade thanks for the heads up, that package looks nicer to use

vlasenko:
Looks like this PR grows bigger and bigger. Maybe it will make sense to split parts of this PR into smaller PRs and merge them. The earlier we start the merge process the easier it will go. When we merge very big PR the kit gets destabilized for days... and we stop every other work and try to stabilize the kit.

verdverm:
I was thinking about how to cherry pick some things. This has become quite a big change set. I was thinking we might do a code review after I get some documentation in place. 

Fortunately the changes are mostly limited to the `user` directory. This has changed to the `auth` directory and the entities directory is new. There are definitely changes to other parts, more so on the frontend.

There are other changes which would probably be good for merging in smaller bites.

- some helpers for 1-1, 1-N, N-N relationships with the new `graphql-resolve-batch` lib (It asks for and fills batches weird, but it is efficient)
- The new authorization method ( pulled and modded from `graphql-auth`) This adds the same withAuth format on the frontend. Should work with the current user/role system
- [I will add more here as I think / find them]

Things that would come later:

- entities
- authz in the DB (keep role as is now)
- config changes and enhancements for entities / authN / authZ
- conditional code for all of this

---

The issue is that `user` and `auth` are present throughout the system and this has been getting large. I'm thinking we start to get some of this pulled into master and then I will cut a new branch and make a swifter, less invasive change for the entities / roles.

verdverm:
The main motivation is two fold.

1. Objects, like posts, should have owners. Owners may like to grant editor permissions to another user for a specific object or set of objects.
1. People organize themselves into groups. Often larger groups have sub-groups.

A couple of major points...

- I've been replacing the generic `id` with more contextual names, this is only on the server/database side so far.
- I've been replacing numeric `id` with `uuid`, this is for security mainly
- I haven't dealt with object ownership ( like posts )
- It may make some sense to do away with the org concept and only have groups, which can also have sub-groups. This would be more flexible and match hierarchical RBAC better

vlasenko:
@verdverm As for `id` -> `uuid` change, I would prefer if it were done through the level of indirection, e.g. you have some function in the kit that generates object id, this function uses `uuid` by default and you can change it for example to use true random seed plus cryptographic random number generator to generate ids.

verdverm:
@vlasenko I was thinking something similar. A configuration option and some implementations. The conversion from int to string IDs was as simple as changing the schema. IIRC, both frontend and backend worked without much of any code change. I've since started using the GraphQL `ID` type in the schemas, which is essentially as string that is given special treatment.

The only thing is, if someone wants to use sequential integers, it would be better to use the `table.increments()` which would make the migration code a little more complex. Maybe integer IDs aren't supported? Just different ID generation methods?

mairh:
Any idea when this will be ready? Also, what are the consequences on existing projects after this is merged to master? I guess this PR is not backward compatible with the existing auth and would break everything.

verdverm:
welp... taking on object ownership necessitates a take II on the permissions scheme

verdverm:
closing, out-of-date, implemented in https://github.com/hofstadter-io/hof-ausk-core-packages

PR #566, opened 2017-12-13
Refactor react comp

mairh:
- React components refactor to utilize new styles.
- Enabled eslint `react/sort-comp` property to ensure the correct order of react functions.

Default is
```
{
  order: [
    'static-methods',
    'lifecycle',
    'everything-else',
    'render'
  ],
  groups: {
    lifecycle: [
      'displayName',
      'propTypes',
      'contextTypes',
      'childContextTypes',
      'mixins',
      'statics',
      'defaultProps',
      'constructor',
      'getDefaultProps',
      'getInitialState',
      'state',
      'getChildContext',
      'componentWillMount',
      'componentDidMount',
      'componentWillReceiveProps',
      'shouldComponentUpdate',
      'componentWillUpdate',
      'componentDidUpdate',
      'componentWillUnmount'
    ]
  }
}
```

### DIFF
--- a/src/client/modules/counter/containers/Counter.jsx
+++ b/src/client/modules/counter/containers/Counter.jsx
@@ -11,6 +11,16 @@ import ADD_COUNTER from '../graphql/AddCounter.graphql';
 import COUNTER_SUBSCRIPTION from '../graphql/CounterSubscription.graphql';
 
 class Counter extends React.Component {
+  static defaultProps = {
+    loading: false,
+    subscribeToMore: () => {}
+  };
+
+  static propTypes = {
+    loading: PropTypes.bool.isRequired,
+    subscribeToMore: PropTypes.func.isRequired
+  };
+
   constructor(props) {
     super(props);
 
@@ -53,11 +63,6 @@ class Counter extends React.Component {
     return <CounterView {...this.props} />;
   }
 }
-
-Counter.propTypes = {
-  loading: PropTypes.bool.isRequired,
-  subscribeToMore: PropTypes.func.isRequired
-};
 
 const CounterWithApollo = compose(
   graphql(COUNTER_QUERY, {
--- a/src/client/modules/counter/containers/Counter.jsx
+++ b/src/client/modules/counter/containers/Counter.jsx
@@ -11,11 +11,6 @@ import ADD_COUNTER from '../graphql/AddCounter.graphql';
 import COUNTER_SUBSCRIPTION from '../graphql/CounterSubscription.graphql';
 
 class Counter extends React.Component {
-  static defaultProps = {
-    loading: false,
-    subscribeToMore: () => {}
-  };
-
   static propTypes = {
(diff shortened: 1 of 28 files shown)


src/client/modules/counter/containers/Counter.jsx
```diff
@@ -11,6 +11,16 @@ import ADD_COUNTER from '../graphql/AddCounter.graphql';
 import COUNTER_SUBSCRIPTION from '../graphql/CounterSubscription.graphql';
 
 class Counter extends React.Component {
+  static defaultProps = {
+    loading: false,
+    subscribeToMore: () => {}
```
mitjade: Do we gain anything with defining this for every function?

mairh: It is just a good practice to define `defaultProps`, especially in the case where the `PropTypes` is marked as required. If the default is not defined it could create problems down the line in the code where you see missing prop `jsx` errors.

vlasenko: @mairh Why would anyone want to set `defaultProps` on `required` props? I can understand why they can be needed on optional props in some cases, but I'm totally don't get why use them on `required` props?

vlasenko: @mairh When defaultProps are used with require props, the typechecking is lost (e.g. React no longer checks if required property was provided), since props typechecking happens after applying defaultProps. See here:
https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values

mairh: So shall we just use it for non required props only?

vlasenko: @mairh I don't think it should be used for any props. Only if you want to have default value for optional prop.

mairh:
Should be good to merge now.

src/client/modules/subscription/containers/UpdateCard.js
```diff
@@ -12,14 +11,6 @@ import settings from '../../../../../settings';
 
 // react-stripe-elements will not render on the server.
 class UpdateCard extends React.Component {
-  static defaultProps = {
-    updateCard: () => {}
-  };
-
-  static propTypes = {
```
mitjade: Here you removed propTypes also?

mairh: Yes because the `propTypes` wasn't used in the component.

mitjade: I is used in `UpdateCardWithApollo`, but I guess it does not count for there, since it is outside of `UpdateCard`.

mairh: I think we discussed this before and queries wrapped under `compose` does not need to have `PropTypes` defined

mitjade: Yes

PR #570, opened 2017-12-14
Contact us component

mairh:
- Contact us page component to receive users query/questions.
- The old email account was expired so I have created a new one.

### DIFF
--- a/.env
+++ b/.env
@@ -19,8 +19,8 @@ GOOGLE_CLIENTSECRET=
 # Email
 EMAIL_HOST=smtp.ethereal.email
 EMAIL_PORT=587
-EMAIL_USER=nxau5pr4uc2jtb6u@ethereal.email
-EMAIL_PASSWORD=7n8dB4pYrrchh452wf
+EMAIL_USER=rwrrrmiyvsgrhocw@ethereal.email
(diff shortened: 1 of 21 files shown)


vlasenko:
Thank you @mairh. Can we make Content field a textarea? Would be great to support attachments too, but this can be left to another PR in the future

mairh:
Good point. It should be `textarea`. I was also thinking about supporting attachments, but then what will happen if someone removes the upload component?

That is why we haven't implemented profile pics upload, adding attachments in the post-comment section.

vlasenko:
Well, I think we can support mechanism for features to announce themselves through connector. For example upload feature can add `upload: true` to the `announcements` object and then `contact` feature can check if `upload` is present in `announcements` and enable or disable attachments based on that?

mairh:
That is one idea - I guess we can do that. Since keeping modularity we are limiting things when it comes about feature-wise. So having a flag for each feature/module and checking if the flag is set to true, we can utilize those modules in another one.

mairh:
@vlasenko Shall we create another PR for profile picture upload and contact us / post-comment attachment addition.

vlasenko:
@mairh Yep, lets do that in separate PRs

PR #573, opened 2017-12-15
Use common client validation

mairh:
- Centralized client validation


mairh:
Is the location of validation.js correct? I am confused between putting it under `common/validation.js` or under `client/helpers/validation.js`

vlasenko:
@mairh I think `common/validation.js` for now

Issue #582, opened 2017-12-20
Modular Role Scope


johnthepink:
I noticed the scope for `graphql-auth` is currently nested under the user module. I thought it would be cool if different modules could declare their own scope and then compose them together. 

Any thoughts on this? I'd be happy to get a PR going. Thanks!

mairh:
That would be a nice addition. I remember we were talking about something like this in the gitter chat.

johnthepink:
@mairh cool I'll do a search through there and take a crack at it.

verdverm:
You may look at what I have been doing on this branch with auth/scopes 
 https://github.com/sysgears/apollo-universal-starter-kit/pull/562 

`config/auth.js` is a good outline of the auth system.

`graphql-auth` has some limitations, I pulled the code into `src/common/authValidation.js` The backend "withAuth" is joined by a "checkAuth" which uses the same permissioning format.

The setup should allow for many numbers of scope styles and verification. I have some concern about the way `graphql-auth` and withAuth does it's validations. (It's basically a pairwise comparison between required scopes and provided scopes that replaces '*' with '.*' and then does a regex check. Seems that could go awry and be difficult to catch security bugs.

Still have to think about how validation works with context and object ownership...

PR #584, opened 2017-12-21
[WIP] Modular Scope

johnthepink:
As per #582, this allows each module to dictate its own scope by passing it along to the Feature instance. Then, the `auth` property is injected in to the context later.

All feedback welcome :)

### DIFF
--- a/src/server/middleware/graphql.js
+++ b/src/server/middleware/graphql.js
@@ -9,10 +9,18 @@ import log from '../../common/log';
 export default async (req, res, next) => {
   try {
     const context = await modules.createContext(req, res);
+    const scopes = modules.createScopes();
+    const { user } = context;
 
     graphqlExpress(() => ({
       schema,
-      context,
+      context: {
+        ...context,
+        auth: {
+          isAuthenticated: user ? true : false,
+          scope: user ? scopes[user.role] : null
+        }
+      },
       debug: false,
       formatError: error => {
         log.error('GraphQL execution error:', error);
--- a/src/server/modules/connector.js
+++ b/src/server/modules/connector.js
@@ -3,7 +3,7 @@
 import type { DocumentNode } from 'graphql';
 import type { Middleware, $Request } from 'express';
 
-import { merge, map, union, without, castArray } from 'lodash';
+import { merge, mergeWith, isArray, map, union, without, castArray } from 'lodash';
 
 import log from '../../common/log';
 
@@ -19,7 +19,8 @@ type FeatureParams = {
   beforeware?: Middleware | Middleware[],
   middleware?: Middleware | Middleware[],
   createFetchOptions?: Function | Function[],
-  catalogInfo: any | any[]
+  catalogInfo: any | any[],
+  scopes: Object
 };
 
 class Feature {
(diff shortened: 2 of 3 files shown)


src/server/middleware/graphql.js
```diff
@@ -9,10 +9,18 @@ import log from '../../common/log';
 export default async (req, res, next) => {
   try {
     const context = await modules.createContext(req, res);
+    const scopes = modules.createScopes();
+    const { user } = context;
 
     graphqlExpress(() => ({
       schema,
-      context,
+      context: {
```
vlasenko: This is a core file of the kit, it should not have as generic as possible. So I would rather continue keeping context shape specifics in relevant modules, in this case in user module.

johnthepink: @vlasenko that makes sense. Do you think modules should be able to provide their own scope? Or, do you think in practice people will just add to the user scope file?

vlasenko: @johnthepink It is likely that I don't quite follow why you moved this out of `user` module into the core. Modules can freely add their data to context, they can add their scopes to context -> auth -> scopes, I guess? Or is there something that I miss?

johnthepink: @vlasenko Thanks for taking the time to help! It's more likely something that I'm missing. :) Let me explain better the issue I'm seeing. I definitely want to come to the right conclusion.

So, `graphql-auth` expects this to exist in the context:
```js
{
  auth: {
    isAuthenticated: bool,
    scope: [] // array of strings saying what the user can view
  }
}
```
On the master branch, that is done in the user module: https://github.com/sysgears/apollo-universal-starter-kit/blob/master/src/server/modules/user/index.js#L174

So you either get the admin scope or the user scope, and then that array is fed in to the auth context object. My thought was that modules would be able to add additional scopes by passing it to the Feature class, merging them all together, and then creating that auth context. However, the user module does not have access to that combined scope from other modules when it creates the auth context. So my thought was to inject the auth context at a higher level after all the Features had been combined.

I hope that makes sense.

vlasenko: @johnthepink Yes, but what stops modules to add their scopes to context in their own `createContextFunc` passed to Feature. All these creteContextFunc execution happens before GraphQL resolvers and `graphql-auth` happens later. We just need to properly merge modules contexts in the connector.js so that they don't overwrite each another and this is the responsibility of connector.js. So my take on this is that we need to have proper merging in our connector.js instead.

vlasenko: @johnthepink Another possibility is that all the modules that depend on `user` module must be imported later in `client/modules/index.js`. We have full control of the order of imports there

verdverm: This is more or less how I am handling auth and modules in my mad sci branch.
The main client/modules/index.js assembles the desired modules in the correct order.
The other client modules import the Auth file with all those `<Auth...>` definitions. Most only use the `<AuthRoute>` which uses a `checkAuth` that mirrors `withAuth` from `graphql-auth`. In actuality, the only module doing anything interesting in the createContext is the one pulling apart the token to fill in the auth user and roles.

In my case though, I have separated the ideas of user and auth into their own modules and further subdivided auth into `authentication` and `authorization`. I also pulled the `withAuth`, `checkAuth` and the `validate` function they use into the common dir at the same level as the client and server dirs. I did this so that the permissioning can be more modular even though the server/client need to be on the same page. I'm imaging different authorization modules living outside of client/server directories.

Thoughts?

johnthepink: Thanks for the feedback all. I will take a further look soon.

vlasenko:
Closing PR because of inactivity

Issue #565, opened 2017-12-11
Documentation for Feature class


aneilbaboo:
The Feature class seems like a key conceptual and organizational unit in apollo-universal-starter-kit. It's used throughout the code base, but there's no documentation.  Adding documentation would make this starter kit vastly more usable. 

Here are some thoughts for an outline:

# Feature Class Documentation
    - Overview
       - Purpose 
       - Client Feature vs Server Feature classes
       - When to create a server feature vs client feature
    - Server Feature
       - Arguments
         - schema
         - createResolversFunc
             - When is it invoked? 
             - What should it return?
             - Using pubsub 
                 - explain the pubsub argument
         - createContextFunc
             - when is it invoked?
             - what should it return?
         - middleware
       - Hierarchical composition
           - explain `new Feature(counter, post, user, graphqlTypes);` in modules/index.js`
    - Client Feature
       - Arguments
          - navItem
          - tabItems
          - route
          - reducer
          - features
       - Hierarchical composition
     - Tutorial:
        - add a server feature
             - add a model using a migration
             - add graphql schema
             - add resolvers
        - add a client feature that uses the server feature

mitjade:
@aneilbaboo Thank you for the outline. I agree that this kind of documentation is of great importance.

Currently we are busy with improving other parts of this kit, but would be really grateful for any kind of help in this area. If you would be willing to start writing docs for some of this parts on your own, we can  help out and fill in the missing parts, when time permits us.

mitjade:
For starters we can create a `doc` directory and write `.md` files there and when something is done, we can link it to the `README.md`. Later we plan to have a separate site for docs.

aneilbaboo:
Hi @mitjade I haven't quite grokked what is going on here myself. I'm a React newbie, so if there's someone else reading this 👋 maybe they can take a crack?  

I'm evaluating a number of different frameworks right now. If I settle in on this repo, I might be able to contribute.

aneilbaboo:
For example, I only have a vague understanding of what the feature class is doing, where it is invoked, when to use it, etc. If someone from the team could fill in at least the overview section, that would be a good start.

mitjade:
@aneilbaboo Thank you for your feedback. Will try and put something together when time permits us.

verdverm:
I can put something in about Features with my upcoming auth docs. Here are some pointers in the meantime:

[...]

PR #591, opened 2017-12-23
Refactor Feature/Module/connector => "Plugin" (Second try)

aneilbaboo:
See discussion at #565

Renamed modules folders to plugins
Renamed connector.js to plugin.js
Renamed Feature class to Plugin
Renamed modules variables to plugins
Renamed misc other Module___.___ files to Plugin___.___


aneilbaboo:
Still some problems here. Will fix and reissue a new PR.

PR #592, opened 2017-12-23
start the damn docs

verdverm:
Pulled some things from the issues and other places, but as the title says...

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,0 +1,62 @@
+# Features
+
+
+- Overview
+   - Purpose 
+   - Client Feature vs Server Feature classes   
+- Server Feature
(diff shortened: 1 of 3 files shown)


vlasenko:
These feature docs doesn't look good to me, maybe it will be better to start documenting simpler things first, then some abstract concepts?

verdverm:
I think my intention here wasn't for quality. Getting a `docs` folder into the mainline will lower the barrier to documentation being written in my experience. Especially with development, where the docs and code can be open in the same editor, copied and pasted, same git workflow. 

What do you think?

vlasenko:
Agreed, merged

PR #593, opened 2017-12-23
Refactor Feature/Module/connector => "Plugin" (Third try)

aneilbaboo:
See discussion at #565

Renamed modules folders to plugins
Renamed connector.js to plugin.js
Renamed Feature class to Plugin
Renamed modules variables to plugins
Renamed misc other Module___.___ files to Plugin___.___

@vlasenko - This is the third try. I was careful to do directory renamings first, followed by file renamings, then alterations of file contents, each as separate commits, but git/Github is treating a couple of graphql files as different. It's weird because neither the filename nor contents changed, just the directory. I've spent a few hours on this taking different approaches, but it's not working.

Can we get this merged?


vlasenko:
@aneilbaboo It cannot be merged, because it is not only couple graphql files, I see many files in the tools dir where Git has lost track of renamings

vlasenko:
In your last commit: https://github.com/sysgears/apollo-universal-starter-kit/pull/593/commits/66d236f27e632188e4d147c37941f162aadef49f
you are doing a bunch of renamings and change the contents of the files at the same time. Of course Git losts track of files because of that

vlasenko:
@aneilbaboo I'm looking to have two commits in this pull request. The first commit does all the renamings. The second commit change all the files contents as needed. Please be careful with that and use `git rebase` to group your commits. You are trying to do non-trivial thing and it should be done carefully

vlasenko:
@aneilbaboo Lets hold off with that. I think I should document current approach carefully first, then we will think about better naming. I plan to do so on the second week of January, as I'm on a vacation right now

vlasenko:
@aneilbaboo Will `plugin` still be a good name if/when we start npm-like repo for them?

vlasenko:
Maybe plugins is a good name. Imagine you have npm-like repo and you install these plugins into your project. To install them you need to have right `socket` in your project that these plugins can fit into. So having `client/plugins/plugin.js` is misleading in this sense, its better to have `client/plugins/socket.js` right?

aneilbaboo:
@vlasenko - I'll hold off on anymore changes. 

I think `plugin` works for something like npm. Grunt has plugins, for example. I'm guessing they use that word to avoid collision with npm's `modules`.  Chrome and Firefox used to have plugins and a site for accessing them. I think they both switched to different formats ("extensions" and "add-ons") in recent versions.

vlasenko:
@aneilbaboo Yes, I think `plugins` should work for us too. But naming as interface as `plugin` in `src/plugins/plugin.js`, might be not very good idea, rather it should be `connector` as it is now, or `socket` or something like that...

PR #594, opened 2017-12-23
graphql-resolve-batch-helpers

verdverm:


src/server/sql/batch.js
```diff
@@ -0,0 +1,86 @@
+import _ from 'lodash';
+
+export const reconcileBatchOneToOne = (sources, results, matchField) => {
```
mairh: Shall we refactor our codebase into a PR to start using GRB instead of Dataloader so that this file is actually being used in the code? At the moment, this is just a dead code sitting in the codebase.

PR #595, opened 2017-12-23
pull common SQL adapters and helpers into reusable functions

verdverm:
…functions

### DIFF
--- a/docs/sql-helpers.md
+++ b/docs/sql-helpers.md
@@ -1,0 +1,93 @@
+# SQL helpers
+
+Files can be found in [src/server/sql](/src/server/sql).
+
+These are functions useful in ETL and reshaping operations
+for database -> resolver data shapes.
+
+### orderedFor
+
+In `helpers.js`
+
+```
+orderedFor( rows, collection, field, singleObject )
+```
+
+__rows__ is the rows returned from the database,
+there are multiple rows for each element of the collection.
+
+__collection__ is the set of items you are getting multiple rows for from the database.
+
+__field__ is the object filed in the rows to group the results by
+and then associate the elements of the collection.
+
+__singleObject__ is a boolean which determines if each collection 
+entry is a single object (1-1 relationship) or not (1-N) relationship.
+
+It returns an array of objects in the first case, and
+an array of arrays of objects in the second case.
+
+---
+
+The adapters are a work in progress and 
+require some changes to how arguments
+to the graphql query are passed around in the server.
+The query also has more parameters, though maybe they should
+be combined into a single object with subfields
+for the different capabilities...?
+
+###  paging
+
+In `paging.js`, adds limit and offset
+
+```
+paging(queryBuilder, args)
+```
+
+__queryBuilder__ is a knex instance.
+
+__args__ is an object with fields `{limit, offset}`
+
+The function returns the queryBuilder object after checking for and adding
+limit and/or offset knex calls.
+
+### cursor
+
+TBD... cursor based paging
+
+
+### currentOrdering
+
+In `ordering.js`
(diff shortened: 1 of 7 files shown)


mitjade:
@verdverm You are doing an amazing job here. I am doing something a bit differently in https://github.com/sysgears/apollo-universal-starter-kit/pull/524. That is generating sql queries with `knex`, based on `resolverInfo`. I think combining both things could resolve into something really powerful. I initially tried integrating `join-monster` into this kit, but it did not fit in well in all the places I needed it. `knex` query builder is really powerful and above all flexible enough to do what you want.

Quick intro to my approach:
In the resolver there is a forth argument `info` for example:
```
product: async (obj, args, context, info) => ...
```
With the help of `graphql-parse-fields` you can get all query fields and then I generate all the select fields and joins so that in the end you get with the help of `knexnest` exact data that you need.
 If you are interested in what I've done, you can check `selectBy` in `https://github.com/sysgears/apollo-universal-starter-kit/blob/cli-crud/src/server/sql/helpers.js`.
To take care of one-to-many relations I needed to use `graphql-resolve-batch`, so I could pass in the `info`. But for this I need to generate a resolver to connect it all together.

Branch is currently work in progress and many things still do not yet fully work as intended.

I am officially on vacation an really promised my wife I will not do any work over the holidays. If any of this interest you we can discuss it further, but I will not really be responsive, maybe twice day at most. After the holidays I will continue on this and I can already see many things you done, that can be useful  to me in what I am doing.

verdverm:
I was definitely thinking about how the query path will probably be needed in the `withAuth` adapter, pairing with the info comes that comes from obj/sources

verdverm:
Check these out: 

https://github.com/sysgears/apollo-universal-starter-kit/blob/96dcc866b91004ffad3d13ff1d811c665e3a8546/src/server/sql/crud.js

https://github.com/sysgears/apollo-universal-starter-kit/pull/562/commits/96dcc866b91004ffad3d13ff1d811c665e3a8546

mitjade:
@verdverm Looks interesting! A bit hard to wrap my head around all the code. Will take a closer look, when I have some time.

verdverm:
I added a HOC Table based on react-table that integrates with the paging crud helper in my madd sci PR

verdverm:
I have a hard time wrapping my head around it too!

src/server/sql/filters.js
```diff
@@ -0,0 +1,135 @@
+import { decamelize } from 'humps';
+import { has } from 'lodash';
+
+export function currentFilter(queryBuilder, filter) {
```
vlasenko: Looks like this function is specific to `user` module. Why is it here, then?

docs/sql-helpers.md
```diff
@@ -0,0 +1,93 @@
...
+
+In `filters.js`
+uses the current code and filter argument
+
+
+### filterBuilder
+
+uses a more advanced filtering input and processing.
+
+See https://github.com/sysgears/apollo-universal-starter-kit/issues/569
```
vlasenko: Is it okay to send user to the PR in the docs?

verdverm:
I've upgraded this system a bunch on my mad-sci branch. Lots of other helpers for other things too. They are starting to become better together. Check the latest commit(s) over there

mairh:
Wouldn't it make sense to pull all those upgrades into smaller PR? mad-sci branch is f*****g huge and kinda impossible to do a PR or make sense out of it.

vlasenko:
@verdverm Are you going to take into account the points I have raised in this PR to finish it and merge?

verdverm:
I should probably cut a new PR on this one too.

PR #609, opened 2018-01-15
Only render on client higher order component

adam-s:
Created a higher order component that prevents a wrapped component from being rendered on the server during server side rendering.


vlasenko:
@adam-s Thank you looks good! One minor thing, higher order component name should start with lower case to not confuse it with ordinary component, because HOC is a function, not a class

src/client/modules/common/components/index.jsx
```diff
@@ -1,3 +1,4 @@
 // eslint-disable-next-line import/prefer-default-export
 export { default as Button } from './Button';
 export { default as LayoutCenter } from './LayoutCenter';
+export { default as ClientOnly } from './ClientOnly';
```
vlasenko: The export name here should be `clientOnly`, case is important. And the file name should be `clientOnly.jsx`

vlasenko:
Looks good now, thank you!

PR #612, opened 2018-01-17
{...this.props} needs to be passed into wrapped component

adam-s:
When I created the HOC I didn't pass this.props into the wrapped component so it breaks on the subscriptions page.


vlasenko:
@adam-s This pull request has conflicts and cannot be merged

adam-s:
This bug breaks the /subscriptions page.

PR #623, opened 2018-01-23
Adding Apollo Link State Functionality (Draft)

sergei-gilevich:
In this pull request i added apollo-link-state to project. Also i removed redux dependency from counter module.


src/client/modules/connector.web.jsx
```diff
@@ -68,6 +70,11 @@ export default class {
     return merge(...this.reducer);
   }
 
+  get resolvers() {
+    console.log('resolvers: ', merge(this.resolver)[0]);
```
vlasenko: Clean up debugging statements pls

src/client/modules/counter/components/CounterView.jsx
```diff
@@ -23,10 +23,10 @@ const CounterView = ({ loading, counter, addCounter, reduxCount, onReduxIncremen
         <Button onPress={addCounter(1)}>Click to increase counter</Button>
         <View style={styles.element}>
           <Text style={styles.box}>
-            Current reduxCount, is {reduxCount}. This is being stored client-side with Redux.
+            Current reduxCount, is {counterState}. This is being stored client-side with Redux.
           </Text>
         </View>
-        <Button onPress={onReduxIncrement(1)}>Click to increase reduxCount</Button>
+        <Button onPress={addCounterState(1)}>Click to increase reduxCount</Button>
```
vlasenko: Text should be updated, it is not Redux counter anymore

src/client/modules/counter/containers/Counter.jsx
```diff
@@ -9,6 +8,8 @@ import CounterView from '../components/CounterView';
 import COUNTER_QUERY from '../graphql/CounterQuery.graphql';
 import ADD_COUNTER from '../graphql/AddCounter.graphql';
 import COUNTER_SUBSCRIPTION from '../graphql/CounterSubscription.graphql';
+import COUNTER_QUERY_CLIENT from '../graphql/CounterQuery_client.graphql';
```
vlasenko: Why the convention change? I thought we have started to use `.client.graphql` in the projects, not `_client.graphql`, why this change?

src/client/modules/counter/containers/Counter.spec.js
```diff
@@ -7,7 +7,7 @@ import COUNTER_SUBSCRIPTION from '../graphql/CounterSubscription.graphql';
 chai.should();
 
 const COUNTER_APOLLO_VALUE = 11;
-const COUNTER_REDUX_VALUE = 7;
+const COUNTER_REDUX_VALUE = 1;
```
vlasenko: Why it is named REDUX here and why you change its value at all? Every change should have a reason.

packages/client/src/modules/counter/containers/Counter.spec.js
```diff
@@ -63,7 +63,9 @@ describe('Counter example UI works', () => {
   step('Increase Redux counter button works', () => {
     const reduxButton = content.find('#redux-button').last();
     reduxButton.simulate('click');
-    content.text().should.has.string(`reduxCount, is ${COUNTER_REDUX_VALUE + 1}.`);
+    setTimeout(() => {
```
vlasenko: This shouldn't be done like that, use `update` call instead

packages/client/src/modules/post/components/PostCommentsView.jsx
```diff
@@ -13,57 +13,55 @@ export default class PostCommentsView extends React.PureComponent {
...
-    subscribeToMore: PropTypes.func.isRequired
+    subscribeToMore: PropTypes.func.isRequired,
+    addCommentClient: PropTypes.func.isRequired
   };
 
   keyExtractor = item => item.id;
 
   renderItem = ({ item: { id, content } }) => {
-    const { comment, deleteComment, onCommentSelect } = this.props;
+    const { comment, deleteComment, addCommentClient } = this.props;
```
vlasenko: Why to call the parameter `addCommentClient`, what does it means? Please use better name here

packages/client/src/modules/post/containers/Post.spec.js
```diff
@@ -327,31 +328,29 @@ describe('Posts and comments example UI works', () => {
...
-
+    editButtons.last().simulate('click');
     const commentForm = content.find('form[name="comment"]');
     expect(
       commentForm
         .find('[name="content"]')
         .last()
         .instance().value
-    ).to.equal('Post comment 2');
+    ).to.equal('Post comment 2'); //must be Post comment 2
```
vlasenko: This comment is redundant. It is evident that it must equal `Post comment 2`, comment doesn't add new information, so it is redundant.

packages/client/src/modules/post/containers/Post.spec.js
```diff
@@ -327,18 +328,17 @@ describe('Posts and comments example UI works', () => {
...
       expect(input.content).to.equal('Edited comment 2');
       done();
       return input;
     };
-
     const editButtons = content.find('.edit-comment');
     expect(editButtons).has.lengthOf(6);
     editButtons.last().simulate('click');
-
+    editButtons.last().simulate('click');
```
vlasenko: Why do we click two times here?

sergei-gilevich: When I use a simulate('click') once, the test is fail. I tried several solutions from this article https://github.com/airbnb/enzyme/issues/1201, but they did not help.
I think this problem is related to the version of the "enzyme".

vlasenko: @sergei-gilevich Oh, I see

PR #625, opened 2018-01-24
Remove Redux dependency in user module

adam-s:
I also removed the Redux Form dependency from the user module using [Formik](https://github.com/jaredpalmer/formik). However, I did not include that in this pull request. I can create another pull request with the code. Also, if perfred I can use [Final Form](https://github.com/final-form/final-form) instead of Formik.


adam-s:
Closing because this is now moved to another branch

PR #626, opened 2018-01-25
Removed Redux dependency from user module

adam-s:


packages/server/src/middleware/website.jsx
```diff
@@ -23,6 +24,7 @@ import Html from './html';
 import Routes from '../../../client/src/app/Routes';
 import modules from '../modules';
 import settings from '../../../../settings';
+import clientModules from '../../../client/src/modules';
```
vlasenko: This is incorrect - it will break working in non-SSR mode. Please return to `require` as it was before.

PR #636, opened 2018-01-31
Integrate "Formik" into the project.

sergei-gilevich:
All `redux-form` forms are replaced with Formik forms in this Pull Request


packages/common/validation.js
```diff
@@ -48,6 +47,9 @@ export const validateForm = (fValues, fSchema) => {
           validateForm1(values[v], schema[v], collector);
         }
       });
+    if (values.password && values.passwordConfirmation && values.password !== values.passwordConfirmation) {
```
evgenijolshanskij: Why is this checker here? What happened when comparable fields have different names?

packages/client/src/modules/post/containers/Post.spec.js
```diff
@@ -338,7 +340,7 @@ describe('Posts and comments example UI works', () => {
     const editButtons = content.find('.edit-comment');
     expect(editButtons).has.lengthOf(6);
     editButtons.last().simulate('click');
-
+    //done();
```
vlasenko: What is this?

packages/common/validation.js
```diff
@@ -28,3 +32,28 @@ export const alphaNumeric = value =>
 // Phone number validation
 export const phoneNumber = value =>
   value && !/^(0|[1-9][0-9]{9})$/i.test(value) ? 'Invalid phone number, must be 10 digits' : undefined;
+
+export const validateForm = (fValues, fSchema) => {
```
vlasenko: why the naming `fValues`, `fSchema` here? should be `formValues`, `formSchema`

packages/common/validation.js
```diff
@@ -28,3 +32,28 @@ export const alphaNumeric = value =>
...
+  let errors = {};
+  const validateForm1 = (values, schema, collector) => {
```
vlasenko: Names must be meaningful, why the name `validateForm1` here, it does not have meaning

vlasenko:
The global question is why only part of kit examples was moved to `Formik`, why not all? Why it is not explained here, what are your intentions?

packages/common/validation.js
evgenijolshanskij: Fixed

packages/common/validation.js
evgenijolshanskij: That is an inner function and not visible from outside, that's why it was called in such way. Anyways, I have renamed it.

vlasenko: @evgenijolshanskij It doesn't matter, names must be meaningful at all times

evgenijolshanskij:
The login form has been left untouched to provide an example of using 'redux-form'. The subscription form is going to be replaced shortly.

vlasenko:
@evgenijolshanskij Lets not include examples of `redux-form`, since there is no anything specific in using `redux-form` with this starter kit

evgenijolshanskij:
@vlasenko Ok, no problem.

PR #642, opened 2018-02-07
enable SSL database connections

claym:
Tested this against postgresql


vlasenko:
@claym Thanks, merged

PR #644, opened 2018-02-07
[WIP] Implement authentication across all platforms (new)

evgenijolshanskij:
This branch was created to replace the outdated 'cross-platform-auth'.


packages/client/src/modules/user/modules/jwt/index.jsx
```diff
@@ -77,34 +91,56 @@ ProfileScreen.propTypes = {
...
-});
-
 export default new Feature({
   tabItem: {
-    User: {
-      screen: LoginNavigator,
+    Profile: {
+      screen: ProfileScreen,
+      userInfo: {
+        requiredLogin: true
```
evgenijolshanskij: Why we cannot just rely on user roles or scopes? Don't you find this approach redundant?

packages/client/src/app/Main.jsx
```diff
@@ -33,10 +33,16 @@ const fetch = createApolloFetch({
...
+  options.credentials = 'same-origin';
+  options.headers = options.headers || {};
+  next();
+});
+
 for (const middleware of modules.middlewares) {
   fetch.batchUse(({ requests, options }, next) => {
-    options.credentials = 'same-origin';
-    options.headers = options.headers || {};
+    // options.credentials = 'same-origin';
```
vlasenko: It is not clear what has happened here and why there is no `TODO`

packages/client/src/app/Routes.jsx
```diff
@@ -1,9 +1,36 @@
...
+    (value.userInfo && value.userInfo.requiredLogin && value.userInfo.role === currentUser.role) ||
+    (value.userInfo && value.userInfo.requiredLogin && !value.userInfo.role) ||
+    !value.userInfo;
+
+  const guestFilter = value => !value.userInfo || (value.userInfo && !value.userInfo.requiredLogin);
+
+  return pickBy(modules.tabItems, currentUser && !currentUserLoading ? userFilter : guestFilter);
+}
+
+class MainScreenNavigator extends React.Component {
```
vlasenko: There certainly should be `TODO` here. Routes should work even if `user` module has been removed from the kit

evgenijolshanskij: They will. If there is no current user logged in 'guestFilter' simply being picked and display all the nav tabs that have not roles restrictions.

vlasenko:
I think we might introduce problems here for users with existing projects to upgrade to new `user` module. Looks like files that work with JWT are not renamed but just readded and users will have hard time to merge these changes

packages/client/src/app/Routes.jsx
vlasenko: @evgenijolshanskij Yes, but it is still not right to have auth-specific code in the core of the kit, it should be removed from here and hooks into connectors/here should be added instead to make the approach generic

evgenijolshanskij: @vlasenko, well, if I got you right the User module should provide its own implementation of the MainScreenNavigator that can be taken instead of the default one.

vlasenko: @evgenijolshanskij Yes, thats correct. I'm not sure how to properly solve this problem yet, need to think

packages/server/src/modules/user/modules/session/auth/crypto.js
```diff
@@ -2,7 +2,7 @@ import crypto from 'crypto';
 import settings from '../../../../../../../../settings';
 
 // Use password-based key derivation function to derive MAC key and encryption key from secret passphrase
-const _deriveSymmetricKey = salt => crypto.pbkdf2Sync(settings.user.secret, salt, 10000, 32, 'sha256');
+const _deriveSymmetricKey = salt => crypto.pbkdf2Sync(settings.user.secret || '', salt, 10000, 32, 'sha256');
```
vlasenko: Why is this change here? We derive secret keys from nothing, this is a security hole?

packages/server/package.json
```diff
@@ -88,11 +88,13 @@
     "knex": "^0.14.2",
     "lodash": "^4.17.4",
     "minilog": "^3.1.0",
+    "mobile-detect": "^1.4.1",
```
vlasenko: Please remove this dependency and use 
```js
import {Platform} from 'react-native';

if (Platform.OS === '...') {
}
``` 
instead.
`react-native` is automatically linked to `react-native-web` on Web front-end and on backend and this approach will work everywhere in our code.

config/user.js
```diff
@@ -1,7 +1,9 @@
 const CERTIFICATE_DEVSERIAL = '00';
-
+const LOCAL_URL = 'exp://192.168.0.155'; //use you local IP
```
vlasenko: Any problems with using information from `Expo.Constants.linkingUri` for `LOCAL_URL` and `MOBILE_APP_URL` ? I certainly don't want any of these defines here. Use `PRODUCTION_URL = `grapql-app://`` please.

packages/server/package.json
sergei-gilevich: I used Platform.OS, but it always return 'web'.

config/user.js
sergei-gilevich: I can't import Expo in this file. I import Expo to the server side, but I received an error message 

```
/home/sergey/development/apollo-universal-starter-kit/node_modules/expo/src/Expo.js:2
import './environment/validate';
^^^^^^

SyntaxError: Unexpected token import
```

vlasenko: I didn't mean to use `Expo.Constants.linkingUri` in this file, but rather in the source code for mobile platforms instead of `LOCAL_URL` and `MOBILE_APP_URL` settings. `LOCAL_URL` and `MOBILE_APP_URL` settings should be removed from here

packages/server/package.json
vlasenko: @sergei-gilevich Yes, for `server` and `web` Platform.OS will return `web`, for `android` it will return `android` and for `ios` it will return `ios`

packages/client/src/modules/user/modules/facebook/containers/FacebookButton.jsx
```diff
@@ -0,0 +1,119 @@
...
+import { View, StyleSheet, Linking, Button, TouchableOpacity, Text, Platform } from 'react-native';
+import { SecureStore, Constants, WebBrowser } from 'expo';
+import { withApollo } from 'react-apollo';
+import PropTypes from 'prop-types';
+import { FontAwesome } from '@expo/vector-icons';
+import CURRENT_USER_QUERY from '../../../common/graphql/CurrentUserQuery.graphql';
+import { withUser, withCheckAction } from '../../../common/containers/AuthBase';
+
+const { protocol, hostname, port } = url.parse(__BACKEND_URL__);
+console.log(url.parse('https://apollo-universal-starter-kit.herokuapp.com/graphql'));
```
evgenijolshanskij: Remove console.log() all over the project

packages/client/src/modules/user/modules/facebook/containers/FacebookButton.jsx
```diff
@@ -0,0 +1,119 @@
...
+let serverPort = process.env.PORT || port;
+
+const facebookLogin = () => {
+  const url = `${protocol}//${hostname}${serverPort ? ':' : ''}${serverPort ||
```
evgenijolshanskij: Consider the following options:
1. ``` let serverPort = (process.env.PORT || port) ? `:${(process.env.PORT || port)}` : `'';` ```
2. ```${serverPort ? ':'+serverPort : ''}```

xabitrigo:
Is there an expected timeframe for merging this pull request? I'm very interested in it.

vlasenko:
@xabitrigo It is merged into master

xabitrigo:
Yay!

PR #646, opened 2018-02-07
Create external-modules.md

verdverm:


vlasenko:
@verdverm I can make `spinjs` build external `yarn link`-ed modules. But it needs some flag  in `package.json` or some other attribute/means of detection to distinguish them from normal npm modules, that should not be transpiled. Normal `yarn linked` third party modules should not be transpiled via Babel, because they might use completely different build system and build concepts.

vlasenko:
@verdverm I think a safe assumption that the `yarn linked` package should be built via `spinjs` is if the package uses `spinjs` and it is a symlink in `node_modules`. If the package uses `spinjs` and installed in `node_modules` normally from `npm-registry`, i.e. it is not a symlink - it should not be built via `spinjs`

verdverm:
@vlasenko that all sounds good and the way it should be. Would docs like this then belong with spinjs? Should this repo have some docs about where spinjs fits in and where to find spinjs docs?

vlasenko:
@verdverm Spinjs is on a different level of abstraction then this kit. It is a build tool, it cares only how to build packages for each platform and what build and what not to build. So the docs from this PR should be here, but the docs of the kit should refer to spinjs docs for sure.

vlasenko:
We use Yarn Workspaces now to write modules, it is easier approach then `yarn link`, so I'm closing this PR

PR #650, opened 2018-02-09
add updated knex helpers

verdverm:


verdverm:
So this is failing on a missing npm package for UUIDs...

It's used here https://github.com/sysgears/apollo-universal-starter-kit/blob/34953c143c3c27c5fb70177d6d0b45c77c162a99/packages/server/src/sql/crud.js#L13

Which was in-line with @vlasenko thought about being able to create an entity with configurable generators.

What do you think about adding `uuid` to the dependencies?

mairh:
We have added all these nice helpers in the codebase, but not using them at all in the actual module sql queries. I don't think it is actually serving any practical purpose if the helpers are just some dead code.

vlasenko:
@mairh I agree

verdverm:
I'm using them [here](https://github.com/hofstadter-io/hof-ausk-core-packages/blob/master/modules/authorization/server/lib/knex/user.js), off mainline

mairh:
What direction are we taking for the kit in the future? Have we agreed on something already that I am not aware of?

https://github.com/hofstadter-io/hof-ausk-core-packages/blob/master/modules/authorization/server/lib/knex/user.js seems to be an extract of some core logic of the kit and this resides somewhere else (not the same repo owner as the core kit).

For anyone who is going to use the kit, I don't see they will have any benefit if the logic stays here in the core kit and the use case is somewhere else.

PR #668, opened 2018-03-06
[WIP] Distinguish between concept of Web Server and concept of GraphQL API Server

vlasenko:
Currently all the code do not distinguish two concepts - the web server that hosts web front-end code and the GraphQL API server. This confuses people who want to have server side rendering, but still want to use external GraphQL API implementation. Another problem is because Web Server endpoint is not specified the most of the code that needs Web Server URL now hardcodes port 3000 during development and uses port from GraphQL API endpoint. 

The goal of this pull request is to separate these two concepts across all the code and remove port 3000 hardcoding.

### DIFF
--- a/.eslintrc.base.json
+++ b/.eslintrc.base.json
@@ -39,6 +39,7 @@
     "__SSR__": true,
     "__PERSIST_GQL__": true,
     "__API_URL__": true,
+    "__WEBSITE_URL__": true,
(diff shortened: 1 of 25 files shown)


vlasenko:
This PR depends on and should be merged after #644

vlasenko:
Merged into #644

PR #677, opened 2018-03-16
Internationalisation

alekseyzadorozhniy:
Add internationalisation on all platforms via `i18next`


packages/client/src/modules/defaultRouter/index.web.jsx
```diff
@@ -1,12 +1,36 @@
 import React from 'react';
-import { Switch } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Switch, withRouter } from 'react-router-dom';
 
 import modules from '../';
-
+import PageLayout from './components/PageLayout';
```
evgenijolshanskij: I think, in order to stick to a common approach it's better to pass it as Feature param with a wildcard.

alekseyzadorozhniy: Good idea. I will do it.

patches/0001-Move-PageLayout-from-components-to-router.patch
```diff
@@ -0,0 +1,597 @@
+From 9c136463dbf9ccd9abe4620b21dc1d5d0c28eb98 Mon Sep 17 00:00:00 2001
```
evgenijolshanskij: Why is it here?

packages/client/src/modules/upload/locales/en/translations.json
```diff
@@ -0,0 +1,12 @@
+{
```
evgenijolshanskij: Where is the module namespace?

packages/client/src/modules/counter/containers/Counter.jsx
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { graphql, compose } from 'react-apollo';
 import update from 'immutability-helper';
-
```
evgenijolshanskij: Just a blank line?

alekseyzadorozhniy: My mistake

packages/client/src/modules/upload/locales/en/translations.json
alekseyzadorozhniy: The namespace is automatically set

patches/0001-Move-PageLayout-from-components-to-router.patch
alekseyzadorozhniy: Will be deleted in the next commit

packages/client/src/app/Main.jsx
```diff
@@ -57,6 +58,14 @@ for (const middleware of modules.middlewares) {
   });
 }
 
+for (const localization of modules.localizations) {
+  // eslint-disable-next-line
+  const resource = require(`@alienfast/i18next-loader!../modules/${localization.ns}/${localization.path}`);
```
vlasenko: We do we do requires here, instead of doing them in their own modules?

patches/0002-Add-router-props-to-container-component.patch
```diff
@@ -0,0 +1,98 @@
+From bf4da52411a04ba1c24c5ab4254088491b0f36eb Mon Sep 17 00:00:00 2001
```
vlasenko: Why these patches are here?

alekseyzadorozhniy: Already removed

packages/client/src/app/Main.jsx
evgenijolshanskij: The idea was to rid modules of redundant logic in favor of merging all the files in one place. Probably this is not the best place for doing this, but language module is.

vlasenko: @evgenijolshanskij Requires - are not a redundant logic. The require specifies what are the dependencies of the module

vlasenko: @evgenijolshanskij We should avoid dynamic requires at all costs, they are poorly supported by various tools

packages/client/src/app/Main.jsx
```diff
@@ -187,10 +196,14 @@ export default class Main extends React.Component {
       modules.getWrappedRoot(
         <Provider store={store}>
           <ApolloProvider client={client}>
-            <ConnectedRouter history={history}>{Routes}</ConnectedRouter>
+            <ConnectedRouter history={history}>
+              <I18nextProvider i18n={modules.i18n}>{Routes}</I18nextProvider>
```
vlasenko: Do you see `modules.getWrappedRoot` 3 lines above this line ?

packages/client/src/testHelpers/Renderer.js
```diff
@@ -187,9 +193,11 @@ export default class Renderer {
     return mount(
       this.withApollo(
         clientModules.getWrappedRoot(
-          <Router history={this.history}>
-            <Switch>{clientModules.routes}</Switch>
-          </Router>
+          <I18nextProvider i18n={clientModules.i18n}>
```
vlasenko: @alekseyzadorozhniy Do you see `clientModules.getWrappedRoot` one line above?

alekseyzadorozhniy: Oh, I think I understand what you mean

PR #682, opened 2018-03-30
Limits website middleware path matching of periods

dcurletti:
The current implementation doesn't allow for path matching on routes such as: `/accounts/register/:token`, where token is a JWT that has a standard period in it.  

By changing the matcher to check for a period (dot) only inside of the path, this allows us to pass params containing periods in the query.


vlasenko:
@dcurletti Good catch! Thank you very much @dcurletti !

PR #683, opened 2018-03-30
Pass currentUser and router into AuthRoute components

mitjade:
pass along router and user and rest props in AuthRoute


vlasenko:
Thank you!

PR #686, opened 2018-04-03
React hot loader

veeramarni:


vlasenko:
Thank you @veeramarni I have tried this PR, but I don't see what we gain with it. The only thing that we can gain with React Hot Loader - is state keeping in UI, I have tried to enter name in contact form and then changed the text of Submit button, the UI was reloaded and all fields of the form were reset. So we are not keeping the state with this code, hence it is not any better then what we have currently with pure HMR. I have tried React Hot Loader multiple times in practice and though I leave a chance that I just don't understand it well, but in my opinion - it is a piece of junk, it is broken in many places by design on many practical paths. They write about this on their website, yet they declare negative attitude to all issue reports - e.g. everything is works and if you have any problem, then it is the problem either with your code or your config. I have never have enough time to deep deeper into guts of RHL and figure out what it actually does and how we can benefit from it, but their attitude is what stops me the most.

vlasenko:
@veeramarni Another important thing I'm thinking all the time in connection to RHL. Do we really need to keep UI state during development, if it will come at a price of reliability of code reload. With Webpack HMR we have reliable code reload and this is more important in my opinion then keeping UI state. I just want to reduce the need in explicit page refresh during development this is what we fight for. If I cannot trust RHL, because it fails in practical situations, why do I need it at all. I need reliability in the first place, and only in the second place it is nice to keep UI state, but with RHL I don't get this. With pure Webpack HMR, though I get reliability, but not keeping state and so be it, it is fine with me

PR #695, opened 2018-04-05
Fixed crash if createFetch is undefined

luckyyang:
This error is caused by delete the upload module, then createFetch will be undefined. Should do the undefined check

### DIFF
--- a/packages/client/src/app/Main.jsx
+++ b/packages/client/src/app/Main.jsx
@@ -69,7 +69,7 @@ const netLink = ApolloLink.split(
   new WebSocketLink(wsClient),
   new BatchHttpLink({
     fetch:
-      modules.createFetch(apiUrl) ||
+      (!!modules.createFetch && modules.createFetch(apiUrl)) ||
       createApolloFetch({
         uri: apiUrl,
         constructOptions: (reqs, options) => ({


vlasenko:
@luckyyang It is better to have this logic inside definition of createFetch in connector, i.e. here:
https://github.com/sysgears/apollo-universal-starter-kit/blob/master/packages/client/src/modules/connector.web.jsx#L33-L35
and here:
https://github.com/sysgears/apollo-universal-starter-kit/blob/master/packages/client/src/modules/connector.native.js#L15-L17

luckyyang:
@vlasenko You mean add logic in connector to prevent `createFetch` is undefined?
 Cause if createFetch is undefined then `modules.createFetch(apiUrl)` will have problem.

vlasenko:
@luckyyang Yes

luckyyang:
Anyway I will add this pr in my project, sometimes createFetch is `undefined` intentionally, then the app will crash

vlasenko:
@luckyyang Just change the code in connectors from:
``` js
    this.createFetch = combine(arguments, arg => arg.createFetch)
      .slice(-1)
      .pop();
```
to:
``` js
    this.createFetch = combine([{createFetch: () => ({})}, arguments], arg => arg.createFetch)
      .slice(-1)
      .pop();
```
In this case if no `createFetch` is defined by any module, the connector will use an empty function, that returns `undefined`, but the function itself will be always defined

vlasenko:
@luckyyang This is the responsibility of connectors to provide sane defaults, not the code that uses  connectors.

luckyyang:
@vlasenko I will cherry pick your change  👍

vlasenko:
@luckyyang Here you go:
https://github.com/sysgears/apollo-universal-starter-kit/commit/c48dbcb70b873e29d88e36316f03dd027bd6b600

PR #714, opened 2018-05-03
Upload fix

mitjade:


packages/server/src/modules/upload/resolvers.js
```diff
@@ -27,7 +35,7 @@ const storeFS = ({ stream, filename }) => {
 };
 
 const processUpload = async uploadPromise => {
-  const { stream, filename: name, mimetype: type, encoding } = await uploadPromise;
```
mitjade: For some reason `filename` and `mimetype` are undefined?

vlasenko: @mitjade Maybe it should be vice versa, e.g. `name: filename`?

mitjade: I just removed the renaming and it works ok, just not sure why this is not working. If `uploadPromise` returns `name` and you want to assign it to `filename`?

vlasenko: It have returned `filename` previously and I needed `name` from it. And it returned `mimetype` and I needed `type` instead to keep backward compatibility with old upload namings

mitjade: So is it ok now as is?

vlasenko: I think yes

PR #730, opened 2018-05-17
Display "No Posts" Message

vsevolodbvd:
This pull request was created to add "No Posts in the db" message if there are no posts. Also, it contains fix for the posts graphql query.


packages/client/src/modules/post/containers/Post.jsx
```diff
@@ -130,7 +130,8 @@ export default compose(
   graphql(POSTS_QUERY, {
     options: () => {
       return {
-        variables: { limit: 10, after: 0 }
+        variables: { limit: 10, after: 0 },
+        fetchPolicy: 'network-only'
```
vlasenko: Why network only fetch policy?

packages/client/src/modules/post/components/PostList.jsx
```diff
@@ -56,12 +56,10 @@ class PostList extends React.PureComponent {
...
 
-  render() {
-    const { loading, posts, loadMoreRows, t } = this.props;
-    const renderItem = Platform.OS === 'android' ? this.renderItemAndroid : this.renderItemIOS;
-    if (loading) {
-      return <Loading text={t('post.loadMsg')} />;
-    } else {
+  renderPosts = () => {
+    const { posts, loadMoreRows, t } = this.props;
+    if (posts && posts.totalCount && posts.edges.length <= posts.totalCount) {
```
vlasenko: Why `posts.edges.length <= posts.totalCount` condition check needed?

Issue #736, opened 2018-05-23
i18n defaultLang config issue


mairh:
There seems to be a slight problem with i18n module. For e.g, if you set `defaultLang: 'en-US'` in the config file the following things happen:

1) Site open up with EN as a selected language by default, which is ok.
2) Switch to RU using the lang selector and the site render the RU translations, which is ok.
3) Refresh the site manually and the site loads up again with EN, whereas it should be RU since we haven't changed the lang selector.

This scenario works fine if you leave `defaultLang` empty.

I think the `defaultLang` config should only set the default language on first site load. Not switch it back everytime the page is refreshed.

PR #737, opened 2018-05-23
Fix problem with defaultLang choosing on page refresh

vlasenko:
Fix problem with defaultLang choosing on page refresh. Move `langList` into config.

Fixes: #736

### DIFF
--- a/config/i18n.js
+++ b/config/i18n.js
@@ -1,4 +1,5 @@
 export default {
   langPickerRender: true,
-  defaultLang: ''
+  langList: ['en-US', 'ru-RU'],
+  fallbackLng: 'en-US'
 };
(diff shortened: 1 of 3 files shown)


config/i18n.js
```diff
@@ -1,4 +1,5 @@
 export default {
   langPickerRender: true,
-  defaultLang: ''
+  langList: ['en-US', 'ru-RU'],
+  fallbackLng: 'en-US'
```
mairh: Changing it to `ru-RU` does not have any effect

vlasenko: @mairh We are matching the list of preferred languages in order of their priority specified in web browser by the user to the list of languages available in our system. Only if that fails we choose `fallbackLng`.

mairh: I see. Since I didn't work on this feature, I wasn't aware that we are detecting user's browser lang pref. I was under the impression that we need to set the default language in the config.

vlasenko:
@alekseyzadorozhniy Thank you Alex!

PR #739, opened 2018-05-24
Mobile chat

Sergiiivzhenko:
Add mobile chat module https://github.com/FaridSafi/react-native-gifted-chat to starter-kit


packages/server/src/modules/chat/resolvers.js
```diff
@@ -3,15 +3,25 @@ const MESSAGES_SUBSCRIPTION = 'messages_subscription';
...
 export default pubsub => ({
   Query: {
-    messages(obj, args, context) {
-      return context.Chat.getMessages();
+    async messages(obj, args, context) {
+      const result = await context.Chat.getMessages();
+      return await result
+        .map(async item => {
+          if (item.userId) {
+            const user = await context.User.getUser(item.userId);
```
evgenijolshanskij: Why can we not fetch users information in one request along with the messages?

packages/client/src/modules/chat/containers/Chat.native.jsx
```diff
@@ -2,57 +2,68 @@ import React from 'react';
...
 @translate('chat')
 class Chat extends React.Component {
   static propTypes = {
-    t: PropTypes.func
+    t: PropTypes.func,
+    messages: PropTypes.array,
+    addMessage: PropTypes.func
   };
 
+  constructor(props) {
```
evgenijolshanskij: Don't create a constructor if you not going to use it. It's unnecessary and just pollutes a class.

packages/client/src/modules/chat/containers/Chat.native.jsx
```diff
@@ -26,6 +42,43 @@ class Chat extends React.Component {
...
+        prev,
+        {
+          subscriptionData: {
+            data: {
+              messagesUpdated: { mutation, node }
+            }
+          }
+        }
+      ) => {
+        let newResult = prev;
```
evgenijolshanskij: Since we stick to a functional paradigm, always try to avoid mutable variables. It is just a matter of a good style.
I know that this is the solution from other modules, but it doesn't mean it is OK.
Here how we can do it without redundant variables:
`
if (mutation === 'CREATED') {
   return AddMessage(prev, node);
}

return prev;
`

vlasenko: @evgenijolshanskij This introduces other problem - returns in arbitrary points of the function which makes code readability worse and is much more error prone than mutable variables because of that. If to rewrite this function to get rid of mutable variables it should be rather done like this:
```
) => (mutation === 'CREATED' ? AddMessage(prev, node) : prev)
```
or at least like this:
```
) => {
  return mutation === 'CREATED' ? AddMessage(prev, node) : prev;
}
```
Generally function should have as few returns as possible and they should be either in the begging of the function (when checking parameters and returning early, though this should be reasonably avoided too) or as the last statement of the function (which is ideal), never in the middle. This is of a higher priority then mutable variables considerations. Mutable variable consideration should be take care too, but without sacrificing returns policies.

vlasenko: @evgenijolshanskij As a general rule we can also stop adding new feature when we see bad code pattern somewhere in the kit and instead switch to hunting for this pattern throughout all the kit code and fixing it, so that our contributors weren't confused about what patterns in the code they should use.

evgenijolshanskij: @vlasenko Everything you said makes complete sense. In this specific code snippet, unfortunately, we cannot use this approach ```) => (mutation === 'CREATED' ? AddMessage(prev, node) : prev)```, because we have at least 3 mutations types.
Probably, for the clarity we could use ```switch``` constuction with returns in every case.

packages/server/src/modules/chat/resolvers.js
```diff
@@ -0,0 +1,144 @@
...
+      pubsub.publish(MESSAGES_SUBSCRIPTION, {
+        messagesUpdated: {
+          mutation: 'UPDATED',
+          id: message.id,
+          node: message
+        }
+      });
+
+      // publish for edit post page
+      pubsub.publish(MESSAGE_SUBSCRIPTION, { messagesUpdated: message });
```
lyzhovnik: Unused publish, should be deleted, because you are using 'MESSAGES_SUBSCRIPTION' for subscriptions. Also useless variable for this action on the top of the file.

packages/server/src/modules/chat/resolvers.js
```diff
@@ -0,0 +1,144 @@
+import fs from 'fs';
+import mkdirp from 'mkdirp';
+import shell from 'shelljs';
+import { createBatchResolver } from 'graphql-resolve-batch';
+
+const MESSAGE_SUBSCRIPTION = 'message_subscription';
+const MESSAGES_SUBSCRIPTION = 'messages_subscription';
+const UPLOAD_DIR = 'public';
+
+const storeFS = ({ stream, filename }) => {
```
lyzhovnik: Looks great, but It's the same func as we are using in the Upload module. As sentence, we can move this functionality in separate file or module in the future.

Issue #720, opened 2018-05-10
Server error responses not translated


mitjade:
In user module, graphql error responses do not get translated.

We need to add i18n support for server side.

vlasenko:
@mitjade PR #743 has just been merged into master

PR #743, opened 2018-05-25
Server error responses not translated

Werter12:
Add i18n support for server side. Issue #720


packages/server/src/i18n/index.js
```diff
@@ -0,0 +1,11 @@
+import i18n from 'i18n';
+
+export default function i18Init() {
```
vlasenko: should be `export default () => {...`

packages/server/src/app.js
```diff
@@ -10,6 +10,7 @@ import websiteMiddleware from './middleware/website';
 import graphiqlMiddleware from './middleware/graphiql';
 import graphqlMiddleware from './middleware/graphql';
 import errorMiddleware from './middleware/error';
+import i18Init from '../src/i18n/index';
```
vlasenko: There is no such thing as `i18`, there is `i18n`, please rename this variable accordingly

packages/server/src/app.js
```diff
@@ -19,6 +20,8 @@ for (const applyBeforeware of modules.beforewares) {
 
 app.use(cookiesMiddleware());
 
+app.use(i18Init());
```
vlasenko: This should be called from within `i18n` module. `i18n` support must be modular. It should not be in the core

packages/server/src/modules/subscription/resolvers.js
```diff
@@ -119,7 +118,7 @@ export default pubsub => ({
           await stripe.customers.deleteSource(stripeCustomerId, stripeSourceId);
         } catch (e) {
           const e = new FieldError();
-          e.setError('subscription', 'Error cancelling subscription.');
+          e.setError('subscription', context.req.headers.cookie, 'user', 'cancelSubscription');
```
vlasenko: We should avoid adding sophisticated usage patterns like this into the kit. I would rather want usage like:
`const e = new FieldError(context.req, 'user');`
`e.setError('subscription', 'cancelSubscription');`

Please think very carefully what usage patterns you introduce to the code base. It is very important

packages/server/src/i18n/utils/Translator.js
```diff
@@ -0,0 +1,23 @@
+import i18n from 'i18n';
+
+export default class Translator {
+  static translate(cookie, moduleName, messageKey) {
+    const langInCookie = cookie.match(/lang=(.*)-/);
```
vlasenko: Avoid manual cookie parsing, when we are relying on the package to do that already - `universal-cookie-express`. Manual parsing is not reliable compared to using package just for this purpose.

packages/server/src/app.js
tv-dm: Fixed.

packages/server/src/app.js
tv-dm: Fixed.

packages/server/src/i18n/index.js
tv-dm: Fixed.

packages/server/src/i18n/utils/Translator.js
tv-dm: Fixed.

packages/server/src/modules/subscription/resolvers.js
tv-dm: Fixed.

config/i18n.js
```diff
@@ -1,5 +1,13 @@
-export default {
+const clientI18n = {
   langPickerRender: true,
   langList: ['en-US', 'ru-RU'],
   fallbackLng: 'en-US'
 };
+
+const serverI18n = {
```
vlasenko: We shouldn't have per-platform config here. Did you see that any of the config files in the kit are per-platform? No, there is no per-platform files, each config is for all platforms. So, don't do it here as well. Reduce the number of config settings to minimum. Why you have `locales` and `langList` here it is the same thing, remove `locales`, use `langList`, etc.

config/i18n.js
```diff
@@ -1,5 +1,13 @@
...
+  locales: ['en', 'ru'],
+  directory: process.cwd() + '/src/modules/i18n/locales',
```
vlasenko: You are adding `directory` config parameter because you have one directory for translations for all the backend. This is wrong. The client-side i18n were made modular, and all the kit is modular. You should have per-module translations as was done for client i18n. It should be easy to remove module from the kit - you just delete directory of the module and remove import of that module. You shouldn't break this mechanism.

config/i18n.js
```diff
@@ -1,5 +1,13 @@
...
+  cookie: 'locale'
```
vlasenko: Its okay to have parameter for cookie name, but you should use this parameter both for client-side i18n and for server-side i18n.

packages/server/src/modules/common/i18n.js
```diff
@@ -0,0 +1,14 @@
+import i18n from 'i18n';
+
+export default (language = i18n.getLocale(), moduleName, messageKey) => {
+  const clientLanguage = language.split('-')[0]; // Supports both options 'en' and 'en-US'
+  const translation = i18n.getCatalog(clientLanguage)[moduleName][messageKey];
+
+  if (!translation) {
```
vlasenko: Exception shouldn't be thrown here, if there is no translation, the translation for fallbackLng should be used, as it is done for client-side i18n. Generally you should study carefully client i18n and make server i18n work as close to it as possible.

packages/server/src/modules/i18n/index.js
```diff
@@ -0,0 +1,16 @@
...
+import settings from '../../../../../settings';
+
+const initMiddleware = async (req, res, next) => {
+  i18n.configure(settings.serverI18n);
+  i18n.init();
+  next();
+};
+
+export default new Feature({
+  middleware: app => {
```
vlasenko: You should use `beforeware`, not `middleware` here.

packages/server/src/modules/i18n/index.js
```diff
@@ -0,0 +1,16 @@
+import i18n from 'i18n';
+
+import Feature from '../connector';
+import settings from '../../../../../settings';
+
+const initMiddleware = async (req, res, next) => {
+  i18n.configure(settings.serverI18n);
```
vlasenko: You should pass the language from cookie here to i18n and don't force developer to pass this cookie for every `translate` call. You are adding dumb work for the developer, avoid this at all costs.

tv-dm: Not quite sure about this solution, since i18n is a global object and applying changes to it will lead to changes for all the clients who requesting the server at the same time.

packages/server/src/modules/common/i18n.js
```diff
@@ -0,0 +1,14 @@
+import i18n from 'i18n';
+
+import settings from '../../../../../settings';
+
+export default (language = i18n.getLocale(), moduleName, messageKey) => {
```
vlasenko: Please remove `language` parameter it is not needed and it adds redundant work for the developer

packages/server/src/modules/i18n/index.js
```diff
@@ -0,0 +1,22 @@
+import i18n from 'i18n';
+
+import Feature from '../connector';
+import settings from '../../../../../settings';
+
+const config = {
+  directory: process.cwd() + '/src/modules/i18n/locales',
```
vlasenko: I have told this earlier - you should have per-module locales, not global locales for the whole backend. Please check how each module has own locales in frontend and do the same for backend

config/i18n.js
```diff
@@ -1,5 +1,6 @@
 export default {
   langPickerRender: true,
-  langList: ['en-US', 'ru-RU'],
-  fallbackLng: 'en-US'
+  langList: ['en', 'ru'],
```
vlasenko: Have you checked frontend after this change? I think you might have broken frontend. Please check that i18n works for frontend as well.

packages/server/src/modules/i18n/index.js
vlasenko: i18n is not a global object - it is a per-request object. Thats why you initialize this object in express middleware for each new HTTP/HTTPS request

vlasenko: Your approach with using global object is wrong, it will have problems in production. You should redesign your approach to something like this:
https://github.com/i18next/i18next-express-middleware

config/i18n.js
vlasenko: The changes in semantics in configuration parameters should be supported by strong argumentation - why this is needed. Because you are forcing many projects based on this kit to swtich to update their configuration. If you don't have strong arguments, please return back original semantics - e.g. `en-US`, instead of `en`, `ru-RU`, instead of `ru`

packages/server/src/app.js
```diff
@@ -13,12 +13,12 @@ import errorMiddleware from './middleware/error';
 
 const app = express();
 
+app.use(cookiesMiddleware());
```
vlasenko: We cannot execute this before beforeware. Beforeware must be executed first, so we need to find other solution.

packages/server/src/modules/user/resolvers.js
```diff
@@ -28,7 +29,7 @@ export default pubsub => ({
         }
 
         const e = new FieldError();
-        e.setError('user', translate(universalCookies.get('lang'), 'user', 'accessDenied'));
+        e.setError('user', i18n.t('user:accessDenied'));
```
vlasenko: This is wrong - you are still using global object, you need to use the object provided to you by i18next-express-middleware in each request

config/i18n.js
```diff
@@ -1,5 +1,6 @@
 export default {
   langPickerRender: true,
-  langList: ['en-US', 'ru-RU'],
-  fallbackLng: 'en-US'
+  langList: ['en-EN', 'ru-RU'],
```
vlasenko: Why is this change here?

packages/client/src/modules/i18n/index.native.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import i18n from 'i18next';
 import { DangerZone } from 'expo';
-import { reactI18nextModule, I18nextProvider } from 'react-i18next';
+import { I18nextProvider, reactI18nextModule } from 'react-i18next';
```
vlasenko: What is the point in changes like these?

packages/client/src/modules/i18n/index.web.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import i18n from 'i18next';
 import LanguageDetector from 'i18next-browser-languagedetector';
-import { reactI18nextModule, I18nextProvider } from 'react-i18next';
+import { I18nextProvider, reactI18nextModule } from 'react-i18next';
```
vlasenko: What is the point in changes like these?

packages/client/src/modules/i18n/index.web.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
...
 
 import Feature from '../connector';
-import { MenuItem, LanguagePicker } from '../../modules/common/components/web';
+import { LanguagePicker, MenuItem } from '../../modules/common/components/web';
```
vlasenko: What is the point in changes like these?

packages/server/src/modules/connector.js
```diff
@@ -1,18 +1,19 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 import type { DocumentNode } from 'graphql';
-import type { Middleware, $Request, $Response } from 'express';
+import type { $Request, $Response, Middleware } from 'express';
```
vlasenko: What is the point in changes like these?

packages/server/src/modules/connector.js
```diff
@@ -1,18 +1,19 @@
...
 
-import { merge, map, union, without, castArray } from 'lodash';
-
-import log from '../../../common/log';
+import { castArray, map, merge, union, without } from 'lodash';
```
vlasenko: What is the point in changes like these?

packages/server/src/modules/i18n/index.js
```diff
@@ -0,0 +1,39 @@
...
+    detection: {
+      lookupCookie: settings.i18n.cookie
+    }
+  });
+  addResourcesI18n();
+};
+
+export default new Feature({
+  beforeware: app => {
+    initI18n();
```
vlasenko: This MUST not be called inside beforeware, as I told before this code must be executed once - on module initialization, not per-request

config/i18n.js
```diff
@@ -1,5 +1,6 @@
...
+  fallbackLng: 'en-EN',
```
vlasenko: Why is this change here?

packages/server/src/modules/i18n/index.js
evgenijolshanskij: Hi, Victor. The ```beforeware``` function is being called once during the App initialization flow by invoking ```applyBeforeware(app);```. In that function we attach a befoware implementation to an app using ```app.use(something)``` construction. That particular ```something``` is being invoked every request. Please, correct me if I'm wrong.

vlasenko: @evgenijolshanskij Yes, I was wrong and you are correct, still I would like to ask to move out `initI18n()` call to the outer level for clarity. E.g.
```
initI18n();

export default new Feature({
  ...
```

evgenijolshanskij: Well, the thing is that we need modules to be initialized by the time i18n is invoked. Only for that reason we have to do it in that place. On the client side, we encapsulate grabbing localizations to the I18nProvider component that is being used after all the initial flow is completed.

packages/client/src/modules/i18n/index.web.js
tv-dm: I ordered dependencies in alphabetical order.

vlasenko:
@tv-dm Looks good to me, thank you Dmitriy!

PR #753, opened 2018-06-15
No loading sign if fetching more posts.

mitjade:
Do not show loading sign if you fetch more posts and some posts are already loaded.


vlasenko:
Thanlk you Mitja!

PR #762, opened 2018-06-29
[WIP] Migrate to Apollo Server 2.0

alekseyzadorozhniy:
This pull request was created to migrate to Apollo Server 2.0

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -62,7 +62,7 @@
     "apollo-link-state": "^0.4.1",
     "apollo-link-ws": "^1.0.3",
     "apollo-logger": "^0.3.0",
-    "apollo-server-express": "^1.3.1",
+    "apollo-server-express": "^2.0.0-rc.5",
     "apollo-upload-server": "^5.0.0",
     "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.6.1",
--- a/packages/server/src/app.js
+++ b/packages/server/src/app.js
@@ -6,8 +6,8 @@ import path from 'path';
 import { isApiExternal } from './net';
 import modules from './modules';
 import websiteMiddleware from './middleware/website';
-import graphiqlMiddleware from './middleware/graphiql';
-import graphqlMiddleware from './middleware/graphql';
+// import graphiqlMiddleware from './middleware/graphiql';
+import getGraphqlServer from './middleware/graphql';
 import errorMiddleware from './middleware/error';
 
 const app = express();
@@ -48,10 +48,13 @@ if (__DEV__) {
     res.send(process.cwd() + path.sep);
   });
 }
+
 if (!isApiExternal) {
-  app.post(__API_URL__, (...args) => graphqlMiddleware(...args));
+  const graphqlServer = getGraphqlServer();
+  graphqlServer.applyMiddleware({ app, path: __API_URL__, cors: corsOptions });
+  // app.post(__API_URL__, (...args) => graphqlMiddleware(...args));
 }
-app.get('/graphiql', (...args) => graphiqlMiddleware(...args));
+// app.get('/graphiql', (...args) => graphiqlMiddleware(...args));
 app.use((...args) => websiteMiddleware(...args));
 if (__DEV__) {
   app.use(errorMiddleware);
--- a/packages/server/src/middleware/graphql.js
+++ b/packages/server/src/middleware/graphql.js
@@ -1,35 +1,54 @@
-import { graphqlExpress } from 'apollo-server-express';
+import { ApolloServer } from 'apollo-server-express';
 import { formatResponse } from 'apollo-logger';
 import 'isomorphic-fetch';
 
-import schema from '../api/schema';
 import modules from '../modules';
+import schema from '../api/schema';
 import settings from '../../../../settings';
 import log from '../../../common/log';
 
-export default async (req, res, next) => {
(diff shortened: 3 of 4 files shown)


packages/server/package.json
```diff
@@ -77,6 +77,7 @@
     "graphql": "0.13.0",
     "graphql-auth": "0.2.2",
     "graphql-iso-date": "^3.3.0",
+    "graphql-playground-middleware-express": "^1.7.1",
```
vlasenko: Why are you adding this package in this PR, it is not related to this PR. If you want to add support for GraphQL playground IDE, it is okay, but it should be done in separate Pull Request

alekseyzadorozhniy: Ok

packages/server/src/middleware/graphql.js
```diff
@@ -1,35 +1,34 @@
...
-    // If createContext decided to finish response, don't pass error downwards
-    if (!res.headersSent) {
-      next(e);
-    }
-  }
+export default () => {
+  const createContext = async (req, res) => await modules.createContext(req, res);
+  return new ApolloServer({
+    typeDefs: schemas.typeDefs,
+    resolvers: schemas.resolvers,
```
Werter12: What is the point of adding typeDefs and resolvers? We could pass executableSchema instead of it.

packages/server/src/middleware/graphql.js
```diff
@@ -1,35 +1,34 @@
...
-      cacheControl: !!settings.engine.engineConfig.apiKey
-    }))(req, res, next);
-  } catch (e) {
-    // If createContext decided to finish response, don't pass error downwards
-    if (!res.headersSent) {
-      next(e);
-    }
-  }
+export default () => {
+  const createContext = async (req, res) => await modules.createContext(req, res);
```
Werter12: How do we get req and res? It seems they are not defined

packages/server/src/middleware/graphql.js
```diff
@@ -1,35 +1,34 @@
...
+  return new ApolloServer({
+    typeDefs: schemas.typeDefs,
+    resolvers: schemas.resolvers,
+    context: async ({ req, res }) => ({ ...(await createContext(req, res)), req, res }),
```
Werter12: How do we get req and res? It seems they are not defined

vlasenko: @Werter12 Check the relevant docs section:
https://www.apollographql.com/docs/apollo-server/v2/migration-two-dot.html#request-headers
But we have `async` / `await` at this line that are not needed, they do nothing in this case.

Werter12: yea, I get it

packages/server/src/middleware/graphql.js
```diff
@@ -1,35 +1,34 @@
...
+    debug: false,
+    formatError: error => {
+      log.error('GraphQL execution error:', error);
```
vlasenko: @tv-dm I'm not sure whether this is still needed with Apollo Server 2.0. It was needed with 1.x

packages/server/src/app.js
```diff
@@ -1,13 +1,11 @@
...
-import cors from 'cors';
-import bodyParser from 'body-parser';
 import path from 'path';
 
 import { isApiExternal } from './net';
 import modules from './modules';
 import websiteMiddleware from './middleware/website';
 import graphiqlMiddleware from './middleware/graphiql';
-import graphqlMiddleware from './middleware/graphql';
+import getGraphqlServer from './middleware/graphql';
```
vlasenko: It is no longer a middleware, so it shouldn't be located in `./middleware` folder

packages/server/src/middleware/graphql.js
vlasenko: There is no point in using `async`/`await` here and there is no point in having `createContext` function, which now does the same as `modules.createContext` function

packages/server/src/middleware/graphql.js
```diff
@@ -1,35 +1,34 @@
...
+        ? formatResponse({ logger: log.debug.bind(log) }, response, options)
+        : response,
+    tracing: !!settings.engine.apiKey,
+    cacheControl: !!settings.engine.apiKey,
+    engine: settings.engine.apiKey
+      ? {
+          apiKey: settings.engine.apiKey
+        }
+      : false,
+    playground: false
```
vlasenko: What does this parameter mean?

packages/server/src/middleware/website.jsx
```diff
@@ -18,8 +18,13 @@ import modules from '../modules';
 import schema from '../api/schema';
 
 let assetMap;
+const playgroundUrl = '/graphiql';
 
 const renderServerSide = async (req, res) => {
+  if (req.url === playgroundUrl) {
+    return;
+  }
+
```
Werter12: We could do this check earlier and not invoke renderServerSide at all.

veeramarni:
@tv-dm Subscription need to be tweaked as shown in https://github.com/apollographql/apollo-server/blob/e514cdfd4c0ff04851cf2fc98d127ef65fdae26b/docs/source/features/subscriptions.md#subscriptions-with-additional-middleware

vlasenko:
@veeramarni Yes, this will be done in next set of pull requests. One PR per feature. The subscriptions are working now, just in an old way.

PR #770, opened 2018-07-08
Fix self referential deleteUser bug

lgants:
The deleteUser method arguments were recently destructured. In the process, the context user was removed due to a naming conflict with the targeted deletion user. Consequently, the expression: `user.id === user.id` always referenced the same user, thereby always raising the error that prevents users from deleting themselves. I also refactored slightly to mimic the pattern of other methods in the file.


vlasenko:
Thank you @lgants Looks good!

PR #776, opened 2018-07-18
Update Contact Module to use <Mutation> element

tv-dm:
This pull request was created to implement <Mutation> component to the contact module instead of graphql function

### DIFF
--- a/packages/client/src/modules/contact/containers/Contact.jsx
+++ b/packages/client/src/modules/contact/containers/Contact.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { graphql, compose } from 'react-apollo';
+import { Mutation } from 'react-apollo';
 
 import ContactView from '../components/ContactView';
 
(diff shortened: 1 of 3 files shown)


packages/client/src/modules/contact/containers/Contact.jsx
```diff
@@ -1,38 +1,34 @@
 import React from 'react';
-import { graphql, compose } from 'react-apollo';
+import { Mutation } from 'react-apollo';
 
 import ContactView from '../components/ContactView';
 
 import CONTACT from '../graphql/Contact.graphql';
 
 class Contact extends React.Component {
```
vlasenko: There is no longer need to declare Contact element as a class, declare it as a closure-function instead.

packages/client/src/modules/contact/containers/Contact.jsx
```diff
@@ -1,38 +1,34 @@
...
   render() {
-    return <ContactView {...this.props} />;
-  }
-}
+    return (
+      <Mutation mutation={CONTACT}>
+        {mutate => {
+          const addContact = async contactInput => {
```
vlasenko: Lets name it `submitContact` instead

packages/client/src/modules/contact/containers/Contact.jsx
```diff
@@ -1,38 +1,34 @@
...
+              if (contact.errors) {
+                return { errors: contact.errors };
+              }
 
-          if (contact.errors) {
-            return { errors: contact.errors };
-          }
+              return contact;
+            } catch (e) {
+              console.log('ERROR', e.graphQLErrors);
```
vlasenko: I don't like this code, lets handle errors here somehow, lets show them after submit in the form as we do when we `login` for example.

PR #791, opened 2018-07-27
Fix "Adding post works" server test error

tv-dm:
Fix error "'totalCount' of 21, but got 20" on the step "Adding post works".


vlasenko:
Brilliant! Many thanks Dmitriy!

Issue #793, opened 2018-07-31
CLI add/delete breaks on index modification


Soup-Flies:
When doing
`yarn cli addmodule <modulename>` or `yarn cli deletemodule <modulename>`
The process breaks at server update because the server file is named `modules/index.ts`
The client file is named `modules/index.js`

Yet in the `tools/cli/module.js` the path points to

> const path = `${__dirname}/../../packages/${location}/src/modules/index.js`;

only allowing the client to work properly

Soup-Flies:
I have a local working solution.
I am checking location for client versus server to dictate which extension the index is given.
If a different approach would be preferred for a PR let me know and I will update.

PR #794, opened 2018-08-01
Fix CLI add/delete module

lyzhovnik:
This pull request was created to fix problem with CLI  "addmodule" and "deletemodule". Error was because index file in modules directory on the server had typescript extension (ts). Now, It does not depend of extension of the file (It can be any extension).
Fixes: #793


vlasenko:
LGTM

PR #795, opened 2018-08-01
Persisted queries

tv-dm:
Add support persisted queries from "Apollo Server 2.0".

### DIFF
--- a/packages/server/src/net.js
+++ b/packages/server/src/net.js
@@ -2,5 +2,5 @@ import url from 'url';
 
 export const serverPort = process.env.PORT || __SERVER_PORT__;
 export const isApiExternal = !!url.parse(__API_URL__).protocol;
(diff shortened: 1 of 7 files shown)


packages/server/src/net.js
```diff
@@ -2,4 +2,5 @@ import url from 'url';
 
 export const serverPort = process.env.PORT || __SERVER_PORT__;
 export const isApiExternal = !!url.parse(__API_URL__).protocol;
-export const apiUrl = !isApiExternal ? `http://localhost:${serverPort}${__API_URL__}` : __API_URL__;
+export const host = 'localhost';
```
vlasenko: What is isApiExternal is true, why it will still be `localhost` ?

vlasenko:
What about updating `README.md` ?

PR #799, opened 2018-08-03
"Apollo Server 2.0" subscriptions

tv-dm:
Add support subscriptions from "Apollo Server 2.0".


vlasenko:
@tv-dm What about hot module replacement support, which was implemented in previous code, but you have completely dropped it from whatever reason now?

And by the way what is about hot module replacement for GraphQL server too. It was implemented in previous code, but you have dropped it, as well

vlasenko:
This question remains:
"And by the way what is about hot module replacement for GraphQL server too. It was implemented in previous code, but you have dropped it, as well"

Issue #798, opened 2018-08-02
Cannot register for user account


QuentinCWG:
Problem: Cannot register using /register.

Error given: 
```{"data":{"register":{"user":null,"errors":null,"__typename":"UserPayload"}},"errors":[{"message":"context is not defined","locations":[{"line":7,"column":5}],"path":["register","errors"],"extensions":{"code":"INTERNAL_SERVER_ERROR"}}]} <= register({"input":{"username":"Test","email":"test@gmail.com","password":"123456789"}})```

vlasenko:
Commited several fixes. Right now registration process goes without error. But after successful Sign Up user is redirected to Sign In form, which is not an optimal flow. But at least now it works as before, so I'm closing this issue.

vlasenko:
@tronin You have different error and hence different issue. Please open new issue and provide reproduction steps.

PR #800, opened 2018-08-04
Redirect issue at RegistrationForm

clodal:
Related to https://github.com/sysgears/apollo-universal-starter-kit/issues/798

- Updated client package > register mutation code at `Register.jsx`

> Commited several fixes. Right now registration process goes without error. But after successful Sign Up user is redirected to Sign In form, which is not an optimal flow. But at least now it works as before, so I'm closing this issue.

Btw i think this registration issue seems to also affect `stable` branch.

### DIFF
--- a/packages/client/src/modules/user/containers/Register.jsx
+++ b/packages/client/src/modules/user/containers/Register.jsx
@@ -1,15 +1,11 @@
-// React
 import React from 'react';
+import { graphql, compose, withApollo } from 'react-apollo';
 
-// Apollo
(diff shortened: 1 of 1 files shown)


vlasenko:
Yes. It was always like this. We should dhow the page that says user registered successfully, to activate his account he needs to follow instructions sent to his email

clodal:
I see... Didn't know as I never saw that page before. Feel free to close this PR out. I may submit another one redirecting to that page another time if time permits. Thanks!

PR #805, opened 2018-08-07
Convert server Upload module into TypeScript

lyzhovnik:
This pull request provides migration server Upload module to TypeScript. Also, was made refactoring on the client side in the Upload module:
- all logic in the views were moved into container;
- implement updating the file list without refetching data from server after removing the file.


vlasenko:
@lyzhovnik Git has lost track of some files. It is because you did both renaming and modification of files in the same commit. Don't do this. Do renamings without any modifications in one commit.

lyzhovnik:
I didn't know it. Thank you for the tip. I will not do it anymore.

vlasenko:
Looks great, thanks

PR #811, opened 2018-08-13
Delete chai-subset lib and move test helpers for the Users module

lyzhovnik:
This pull request removes chai-subset library, also move test helper file with functions for log in/out to the Users module.


vlasenko:
Thank you Nikita, looks good!

PR #813, opened 2018-08-13
Convert debug, graphqlTypes, i18n and mailer modules to Typescript

lyzhovnik:
This pull request converts debug, graphqlTypes, i18n and mailer modules on the server side to Typescript.


vlasenko:
Awesome, thank you Nikita!

PR #816, opened 2018-08-14
Convert the Post module into Typescript

lyzhovnik:
This pull request provides migration the Post module from JavaScript into Typescript.


vlasenko:
LGTM

PR #818, opened 2018-08-16
Remove playground middleware

lyzhovnik:
This pull request removes Playground middleware because Apollo provides Playground out of the box.
Now Playground is running on the "/graphql" url instead of previous url "/gplayground".
According to the docs: https://www.apollographql.com/docs/apollo-server/features/graphql-playground.html


vlasenko:
Awesome! Thank you very much Nikita!

PR #819, opened 2018-08-20
Convert the middleware folder into typescript

lyzhovnik:
This pull request converts the middleware folder into typescript. Also fixes bug with invalid urls in the error middleware.


vlasenko:
Looks good, thank you Nikita!

PR #828, opened 2018-08-27
Create class for saving/deleting files in the file system

lyzhovnik:
Move logic for saving/deleting the files in the file system into separate class to make it reusable in other modules.


vlasenko:
Thank you Nikita, looks good!

PR #837, opened 2018-08-31
Copyedit and update README

AndSviat:
Copyedited and updated README; moved various sections to Wiki.


vlasenko:
Awesome! Many thanks Svyat!

PR #861, opened 2018-09-26
[WIP] Move docs from Wiki into readme.md files

lyzhovnik:
This pull request moves documentation from the Wiki to the readme.md files. This will provides us to keep one source of truth of the documentation.


packages/server/.env
```diff
@@ -21,5 +21,6 @@ EMAIL_USER=rwrrrmiyvsgrhocw@ethereal.email
 EMAIL_PASSWORD=hMnwJM2Zuwhr6ZrpHe
 
 # Stripe
+STRIPE_PUBLIC_KEY=
```
vlasenko: STRIPE_PUBLIC_KEY is needed only on front-end side, why do you add it here? These settings are read by the backend, not frontend

AndSviat:
The docs are updated according to the latest suggestions.

PR #864, opened 2018-09-27
Password min lenght as a configuration parameter.

mitjade:


vlasenko:
lenght need to be renamed to length everywhere

vlasenko:
Thanks! Merged

PR #867, opened 2018-09-29
[WIP] Fractal Objects architecture for modules

vlasenko:
Utilize [fractal-objects](https://github.com/sysgears/fractal-objects) library for application modules design.


mitjade:
@vlasenko Looks great!

mairh:
Looks good. However, could you tell me the reason for this change? Is it just to have a properly supported open source library instead of having our own logic? Or does this change solve anything new?

vlasenko:
@mairh The approach to make application module resemble fractals that can be easily divided and recombined is not specific to the kit. It can be applied to other projects, including legacy projects to make them more modular and better readable, testable and reusable thanks to this. 

The approach can also be ported to the other programming languages and thanks to this can be further tested and matured. Because of this it makes sense to me to split it into separate self-contained library with tests and documentation (coming soon).

PR #870, opened 2018-10-01
Improve deployment and CLI for Stripe Subscription Module

lyzhovnik:
This pull request provides such works:
-  improves deployment of Stripe Subscription Module to Heroku - do not need to run cli manually, It will run automatically;
- improves the CLI to init the Stripe with the basic product and plan -  creates a new product/plan only if it hasn't been created before, or has different parameters;
- removes private Stripe data from the config file.


vlasenko:
LGTM, thank you Nikita!

PR #874, opened 2018-10-05
Fix check for skippedItems in mobile navigation

theWeird:


vlasenko:
@theWeird Can you explain the difference because your code and previous code. Where was the problem in previous code?

theWeird:
Variable items points to an array of strings (drawer names), extracted by Object.keys(modules.drawerItems)

items[name].skip would try to check for example items["Counter"].skip.
Because items is a numerical array and no object, this causes an error.

Instead, I just use the name variable to check for the skip property in the original drawerItems object.

vlasenko:
@theWeird Thanks, looks good, though I believe `!!` is not needed here

theWeird:
Just to convert null, undefined, etc. from falsy into a real boolean

vlasenko:
@theWeird I understand, what will it give you in this case, if you convert undefined, null into "real" boolean?

theWeird:
Habit from our project. We have the convention to work type-safe.
E.g. convert truthy/falsy to real booleans, use type-save compairsions (===)

So if you don't mind it, there shouldn't be a drawback in omiting it.

vlasenko:
@theWeird Okay, I have already pushed the commit that is equivalent to yours:
https://github.com/sysgears/apollo-universal-starter-kit/commit/940e2160f0f7f3cc5996a18237ab79235ccf682b
And another commit to fix mobile frontend:
https://github.com/sysgears/apollo-universal-starter-kit/commit/abac423f3e306dfbae29ac90aae35bc00c42774d

So, I think that this PR can be closed. I'm sorry for intervention, I was working on fixing mobile frontend too during last hour.

PR #880, opened 2018-10-10
fix twitching of right menu in antd

mitjade:
`style={{ display: 'flex', alignItems: 'center' }}` on `MenuItem` causes the menu to twitch on antd.
I moved this to `class`, so it is only defined for bootstrap.


vlasenko:
Looks good, thank you Mitja!

Issue #882, opened 2018-10-12
Fix tslint Errors for unused vars, if you create a module through the cli


mitjade:
**Describe the bug**
Fix tslint Errors for unused vars, if you create a module through the cli

**To Reproduce**
1. Go to terminal
2. Type `yarn cli addmodule testModule`
3. Run `yarn watch`

**Expected behavior**
No errors should be seen if you create a new module through cli

mitjade:
Should be fixed by changing the template code, so that there are no unused variables. Add to each module trivial SQL query, like SELECT 1 and trivial GraphQL resolver that calls this query and add test that checks it.

PR #883, opened 2018-10-12
Temporary fix for unused var errors.

mitjade:
Quick fix for #882.


mitjade:
Looks good to me @lyzhovnik!

PR #887, opened 2018-10-18
Fix wrong formik handlers defined on SSR.

mitjade:
During the removal of `react-native-web`, this bug was introduced.

Should be fixed by checking the `PLATFORM` instead of `document`, witch environment is active.


vlasenko:
Many thanks Mitja!

PR #893, opened 2018-10-24
Strongly typed resolvers

kamilkisiela:
- [ ] update to `v0.12.7` of codegen

Every resolver is now strongly typed, based on server's GraphQL Schema.

Without types, there's no validation of what resolver returns, what arguments it has and how context looks like.

```ts
{
  Query: {
     async posts(_, { limit, after }, context) {
       return context.Posts.all({limit, after});
     }
  }
}
```

Thanks to GraphQL Code Generator and its TypeScript template we generate types based on schema, for each type, enum, interface etc and also for each resolver.

Now, when a different object is return, different that you schema defines, we get a typescript error.
Because everything is strongly typed, we get autocompletion, instant validation and things like that.

Let me know what do you think :) I'm going to answer any question you have. I know it's a big PR, a lot of things changed but not that many actually, I only added types, no change in the code in terms of logic but feel free to message me so we could review it together.

- GraphQL Code Generator website: https://github.com/dotansimha/graphql-code-generator


vlasenko:
@kamilkisiela Thank you! This comes at a very high price - hot reloading / live editing will no longer work for graphql schema files. Because now after each change to schema files you must manually execute code generator script.

vlasenko:
@kamilkisiela Yep, it has. Though not very convenient, e.g. watchman is needed. Maybe it would be more convenient to write Webpack plugin? I have written Webpack plugin with similar mechanics in the past, it scanned graphql files in webpack modules tree and extracted graphql queries from there. I think this mechanics is very close to what is needed (but we need to not extract queries but rather generate types):
https://github.com/sysgears/persistgraphql-webpack-plugin
What are your thoughts about Webpack plugin approach?

kamilkisiela:
@vlasenko By generating types with webpack, an IDE won't be able to see them so no autocompletion and type checking, it will only show errors in build time.

kamilkisiela:
@vlasenko I thought you meant virtual filesystem for that but yeah it makes sense :) I'll let you know when it's ready then!

vlasenko:
@kamilkisiela Yep, virtual fs won't work for this case. I have actually started looking myself how to make a webpack plugin for this task. So either you or me will hopefully come up with the solution or exchange ideas :)

packages/server/src/modules/user/access/session/resolvers.ts
```diff
@@ -0,0 +1,22 @@
...
+interface Context {
+  req: any;
+}
+
+export default (): {
+  Mutation: models.MutationResolvers.Resolvers<Context>;
+} => ({
+  Mutation: {
+    async logout(obj, args, { req }) {
+      const session = { ...req.session };
```
evgenijolshanskij: Simply, cleaner and without mutability:
```
const { userId, ...session } = req.session;
return writeSession(req, session);
```

packages/server/src/modules/user/resolvers.ts
```diff
@@ -2,22 +2,36 @@
...
 
+import * as models from '../../../typings/graphql';
 import FieldError from '../../../../common/FieldError';
 import settings from '../../../../../settings';
 
 const USERS_SUBSCRIPTION = 'users_subscription';
 
-export default pubsub => ({
+interface Context {
+  User: any;
```
evgenijolshanskij: The ``User`` variable typing is needed, since we add it in other modules

tools/templates/module/server/resolvers.ts
```diff
@@ -1,4 +1,6 @@
-export default (pubsub: any) => ({
+import { PubSub } from "graphql-subscriptions";
```
evgenijolshanskij: ```"graphql-subscriptions"``` - according to a common code style, single quotation marks are required

kamilkisiela:
@evgenijolshanskij done

vlasenko:
@kamilkisiela I saw that you have created Webpack plugin for code generator. Are you going to utilize this plugin here in this pull request?

vlasenko:
@kamilkisiela Closing this PR, as there is no response from the author. I will be happy to include code generation and strongly typed resolvers if you will be willing to cooperate on improving code generator integration with webpack

PR #906, opened 2018-11-13
Convert the Contact module into TypeScript [Client]

lyzhovnik:
This pull request converts the Contact module into Typescript and also contains some addition functionality, such as: 
* Implements validation incoming data on the server side with the same schema which is used on the client side;
* Adds functionality to FieldError class for transforming validation error messages for sending them via Graphql. (simple object transforms into array of objects on the server side, and transforms back on the client side).
* Moves all validation functionality into separate folder in the root modules from common folder


packages/common/utils.ts
```diff
@@ -78,6 +78,18 @@ const getPlatform = () => {
   }
 };
 
+/**
+ * Transforms errors from object into array of object (For Grapqhl type FieldError)
+ */
+export const transformValidationMessagesForGraphql = (errors: { [key: string]: string }) =>
```
lyzhovnik: I think this is not the best name for this function. Does anyone have an idea ?

vlasenko:
Looks awesome. Many thanks Nikita!

PR #915, opened 2018-11-22
Correctly order packages

mitjade:
If you add a new package to server or client, some of the packages get reordered.


vlasenko:
LGTM, thank you Mitja!

PR #917, opened 2018-11-23
Use 'foo-bar-baz' for package module name.

mitjade:
Switch from camelCase to 'foo-bar-baz' for package name in `package.json`


vlasenko:
Thanks!

PR #919, opened 2018-11-23
Add package name to `package.json`

mitjade:
When adding new module, add name to `package.json`.


vlasenko:
@mitjade Looks great, thank you Mitja!

PR #920, opened 2018-11-23
Update imports in website middleware

alekseyzadorozhniy:
Add imports in website middleware only when SSR is enabled


vlasenko:
Thank you Alex, looks good!

PR #921, opened 2018-11-23
[WIP] Implement initial multi-module architecture [Scala]

kmusienko:
This pull-request is intended to define a multi-module SBT architecture of Scala implementation of the Kit.

All the custom modules will reside on the top-level in the `modules` folder. Scala implementation of a module will reside on a `server-scala` subfolder, e.g. `modules/user/server-scala`.

The main SBT module will reside on `packages/server-scala` folder and is intended to run all the modules as a single application.

The 'core' module will reside under the main module in `packages/server-scala/core` and is intended to store the base essential code that will be used by other custom modules.

In this pull-request, I made the 'upload' and 'user' modules to be separate SBT modules and moved them to the top-level `modules` folder. Also, I added configuration that allows running all the SBT modules as a single application.


vlasenko:
The core module is just the same module as the others, why to keep it separately?

kmusienko:
@vlasenko, yes, indeed, I was thinking about it. I will move the core module to `modules/core/server-scala` so all the modules will be kept together.

PR #924, opened 2018-11-23
Rename module-common-react to module-common

alekseyzadorozhniy:
Rename `module-common-react` to `module-common` since the module does not contain anything specific to  React and can be used in other non-React modules


vlasenko:
Yes, indeed. Good catch!

PR #928, opened 2018-11-26
Add user authentication with JWT [Scala]

serjorez:
Implement registration and login with JWT tokens.


modules/core/server-scala/src/main/resources/application.conf
```diff
@@ -1,3 +1,8 @@
+app {
```
tsione: typo in app name

modules/core/server-scala/src/main/resources/application.conf
```diff
@@ -46,4 +52,6 @@ email {
 
 jwt {
   secretKey = "676918AB4D29BFE59CCB943F3C09F5CC8FB3A8511E23E502B67DE95AB9A9D00C"
+  accessTokenExpiration = 1d
```
tsione: maybe we can use smaller expiration time

modules/core/server-scala/src/main/scala/common/actors/Dispatcher.scala
```diff
@@ -20,7 +20,7 @@ object Dispatcher extends ActorNamed {
                                      context: UserContext,
                                      replyTo: ActorRef,
                                      resolverActor: ActorRef,
-                                     onException: Exception => Any,
+                                     onError: Throwable => Any,
```
tsione: should be an exception

modules/core/server-scala/src/main/scala/common/config/AppConfig.scala
```diff
@@ -0,0 +1,9 @@
+package common.config
+
+import com.google.inject.Inject
+import com.typesafe.config.Config
+
```
tsione: should be a singleton

modules/core/server-scala/src/main/scala/common/errors/AlreadyDone.scala
```diff
@@ -0,0 +1,3 @@
+package common.errors
+
+case class AlreadyDone(msg: String = "") extends Error(msg)
```
tsione: should we use another name

PR #930, opened 2018-11-26
React counter update

alekseyzadorozhniy:
Update react counter to use the counter common module


vlasenko:
Great, many thanks Aleksey!

PR #963, opened 2018-12-18
Implement authentication module [Scala]

serjorez:
Move an authentication functionality from user module into a separated module to follow single responsibility architecture.


modules/user/server-scala/src/main/scala/graphql/schema/UserSchema.scala
```diff
@@ -2,201 +2,42 @@ package graphql.schema
...
-    Field(
-      name = "resetPassword",
-      fieldType = resetPayload,
-      arguments = List(Argument("input", resetPasswordInput)),
-      resolve = sc => resolveWithDispatcher[ResetPayload](
-        input = sc.args.arg[ResetPasswordInput]("input"),
-        userContext = sc.ctx,
-        namedResolverActor = UserResolver
-      )
+      resolve = sc => Some(User(Some(1), "user", "test@test.test", "123", "user", true))
```
kmusienko: should return None for now

Issue #858, opened 2018-09-21
Support File Upload for mobile


theWeird:
So far the upload component is only supported for web clients with a small placeholder component for mobile apps.

Would be great to have this feature on mobile platforms as well.
Maybe with two examples, one opening a file chooser and one opening the camera.

RomanPeskov:
Thanks, that's a good idea. We have already implemented uploading files for mobile platforms with file chooser in the mobile-chat branch. So, in the near future we plan to add this functionality for mobile platforms in the upload module.

PR #965, opened 2018-12-19
Implement file upload for mobile platform

RomanPeskov:
This pull request implements file upload for mobile platforms. Also it fixes the [#858](https://github.com/sysgears/apollo-universal-starter-kit/issues/858) issue.


modules/upload/client-react/containers/FileOperations.jsx
```diff
@@ -0,0 +1,33 @@
...
+export default class FileOperations extends React.Component {
+  state = {
+    error: null
+  };
+
+  handleUploadFiles = async files => {
+    const { uploadFiles } = this.props;
+    const result = await uploadFiles(files);
+
+    this.setState({ error: result && result.error ? result.error : null });
```
alekseyzadorozhniy: Perhaps it is better to use a logical OR (||) in similar constructions

modules/upload/client-react/containers/Upload.jsx
```diff
@@ -81,13 +98,18 @@ export default compose(
                 __typename: 'File'
               }
             },
-            update: store => {
-              const cachedFiles = store.readQuery({ query: FILES_QUERY });
+            update: ({ caches }) => {
```
vlasenko: The destructuring assignment here `{ caches }` tries to access private field of a class, `caches` is a private field. This is forbidden. We must not use private fields. Use public API of the class instead

Issue #812, opened 2018-08-13
Get rid of 'updateQueries' mutation methods


lyzhovnik:
The mutation method 'updateQueries' should be replaced  to the method 'update',  because 'updateQueries' method is gonna be deprecated - https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-mutation-options-updateQueries 

Good example for 'update' method implementation you can see in the Upload module.

alekseyzadorozhniy:
The issue has been resolved in the #969 PR

PR #969, opened 2018-12-20
Replace old 'updateQueries' mutation functions with the new 'update' ones

BaloPhilip:
The mutation method 'updateQueries' should be replaced with the 'update' one, because 'updateQueries' method is gonna be deprecated. Also, it fixes the #812 issue.


vlasenko:
Thank you Phillip, looks good!

Issue #968, opened 2018-12-20
Add support for creating modules in the root `modules` dir


vlasenko:
We should update CLI to support creating and removing modules from root `modules` dir, while still supporting creating and deleting modules from `packages/whatever/src/modules`

evgenijolshanskij:
The CLI has been updated in the #970 PR.

PR #970, opened 2018-12-20
CLI modules management for the new structure

evgenijolshanskij:
This PR contains improvements that allow to manage modules in the new file structure via CLI commands (rel. #968).


tools/cli/commands/installDependency.js
```diff
@@ -0,0 +1,17 @@
+const chalk = require('chalk');
+const { spawn } = require('child_process');
+
+function installDependency({ logger, moduleName }) {
```
vlasenko: Why do we run `yarn` when we add module? It takes much time to run yarn. It was not needed before, why do we need it now? Why making symlink is not enough?

evgenijolshanskij: Looks like it is not enough, since I get the error continuously ```spin error Error: Cannot find module '@module/example-client-react'``` after the module is added and `yarn watch` command run.

evgenijolshanskij: Could you please test on your side as well?

vlasenko: @evgenijolshanskij Ok, I will check and report back

Issue #901, opened 2018-11-07
Ignore Case when searching users 


abhayca:
**Describe the issue**
Currently, the comparison when searching for a particular set of users is case sensitive. Ideally speaking, it should be case insensitive.

**To Reproduce**
Steps to reproduce the behavior in the mobile app:
1. Go to Users option
2. Click on Filter icon at the top right
3. In the Search field, enter the string, Admin (note A should be capital)
4. Users Not Found is displayed.

**Expected behavior**
The admin user should be displayed, 

**Answer**
Instead of knex 'like', use 'ilike' in the /server/.../user/sql.js file as follows:

```
async getUsers(orderBy, filter) {
      ...
      if (has(filter, 'searchText') && filter.searchText !== '') {
        queryBuilder.where(function() {
          this.where('u.username', 'ilike', `%${filter.searchText}%`)
            .orWhere('u.email', 'ilike', `%${filter.searchText}%`)
            .orWhere('up.first_name', 'ilike', `%${filter.searchText}%`)
            .orWhere('up.last_name', 'ilike', `%${filter.searchText}%`);
        });
      }
    }
    return camelizeKeys(await queryBuilder);
  }
```
At least this solution works for the PostgreSQL db, I don't know about the other dbs.

evgenijolshanskij:
The issue resolved in the #972 PR.

PR #972, opened 2018-12-21
Improve user search by removing case sensitivity

basilalex:
This pull-request fixing case sensitivity for user search field described in issue [#901](https://github.com/sysgears/apollo-universal-starter-kit/issues/901).


vlasenko:
Will this approach work for most SQL databases, or is it SQLite-specific?

PR #973, opened 2018-12-21
Add Angular Material

alekseyzadorozhniy:
Add Angular Material library as default UI toolkit for client-angular


vlasenko:
Thank you Alex, looks great!

Issue #849, opened 2018-09-14
Knex Transaction when inserting/editing a user


coooolers:
Server Side User Module Resolver.js
1. addUserMutation
`
const [createdUserId] = await User.register({ ...input });
await User.editUserProfile({ id: createdUserId, ...input });
`

2.  editUser Mutation
`
await User.editUser(userInfo);
 await User.editUserProfile(input);
`

These two operations should occur in a transaction in sql.js.

This could be a probable answer:
https://stackoverflow.com/questions/22957032/knex-transaction-with-promises

Regards

coooolers:
This ES7 async/await based is a nicer solution:
https://stackoverflow.com/questions/40580674/get-knex-js-transactions-working-with-es7-async-await

lyzhovnik:
@coooolers Hi, Thank for your notice. Yes, I would be great to implement this via transaction. We'll implement this in the future.

lyzhovnik:
Hi, I have example of using transaction from out [chat](https://gitter.im/sysgears/apollo-fullstack-starter-kit). 
I just save it there fro the future: 

`Create Thing Resolver`
```javascript
const createdThingId = await Thing.createThing({ ...input });
```
`Create Thing SQL`
```javascript
async createThing({ userId, itemId, wishingToDonate, lookingForAid, attributes }) {

    //First you need to Promisify the knex.transaction function. If you don't use the .catch(reject) clause, you'd get an uncatchable unhandled rejection error on rollback.
    const promisify = (fn) => new Promise((resolve, reject) => fn(resolve).catch(reject));
    const trx = await promisify(knex.transaction);

    try {
      const [thingId] = await returnId(trx('thing')
      .insert({ user_id: userId, item_id: itemId }))

      //This loop creates multiple records in the attribute table one by one. The attributes is an array which is a collection of objects where each object is of the form: {name: 'Book Name', value: 'Harry Potter'}

      for (let i=0; i<attributes.length; i++) {

        const attribute = attributes[i];

        //It returns the id of the newly created attribute.
        const [attributeId] = await returnId(trx('attribute')
        .insert({ name: attribute.name, value: attribute.value, thing_id: thingId }));
      }

      await trx.commit();
      return thingId;

    } catch (e) {
      await trx.rollback();
      //await trx.rollback(new Error('Something went wrong'))
    }
  }
```
`Edit Thing Resolver`
```javascript
 editThing: withAuth(
[...]

PR #977, opened 2018-12-24
SQL transactional operations support

StKostyantin:
This pull request add knex transaction for inserting and editing user. Also it fixes the #849 issue.


modules/user/server-ts/sql.js
```diff
@@ -163,13 +161,23 @@ class User {
     return returnId(knex('auth_linkedin')).insert({ ln_id: id, display_name: displayName, user_id: userId });
   }
 
-  async editUser({ id, username, email, role, isActive, password }) {
+  editUser({ id, username, email, role, isActive, password }) {
     let localAuthInput = { email };
     if (password) {
-      const passwordHash = await bcrypt.hash(password, 12);
-      localAuthInput = { email, password_hash: passwordHash };
+      console.log('editUser password');
```
evgenijolshanskij: console.log shouldn't be in the final version

modules/user/server-ts/sql.js
```diff
@@ -163,13 +161,23 @@ class User {
...
+      bcrypt.hash(password, 12).then(passwordHash => {
+        localAuthInput = { email, password_hash: passwordHash };
+
+        return knex('user')
+          .update({
```
evgenijolshanskij: Use decamelizeKeys() here

modules/user/server-ts/sql.js
```diff
@@ -163,13 +161,23 @@ class User {
...
   }
 
-  async editUser({ id, username, email, role, isActive, password }) {
+  editUser({ id, username, email, role, isActive, password }) {
     let localAuthInput = { email };
     if (password) {
-      const passwordHash = await bcrypt.hash(password, 12);
-      localAuthInput = { email, password_hash: passwordHash };
+      console.log('editUser password');
+      bcrypt.hash(password, 12).then(passwordHash => {
```
evgenijolshanskij: Prefer using `await` over `.then()`

modules/user/server-ts/sql.js
```diff
@@ -180,13 +188,15 @@ class User {
       .where({ id });
   }
 
-  async editUserProfile({ id, profile }) {
-    const userProfile = await knex
+  isUserProfile({ id }) {
```
evgenijolshanskij: Why did you decided to take this function out and make separate? Also, the function name doesn't correlate with what it does.

modules/user/server-ts/sql.js
StKostyantin: fixed

modules/user/server-ts/sql.js
StKostyantin: fixed

evgenijolshanskij: Simplify: `{ email, password_hash: passwordHash }` -> `{ email, passwordHash }`

modules/core/common/createTransaction.ts
```diff
@@ -0,0 +1,37 @@
+import { Transaction } from 'knex';
+
+export default async () => {
+  const { knex } = require('@module/database-server-ts');
```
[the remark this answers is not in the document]
StKostyantin: fix

modules/database/server-ts/sql/createTransaction.js
```diff
@@ -0,0 +1,5 @@
+import { knex } from '.';
+
+export default async () => {
+  return await new Promise(resolve => knex.transaction(resolve));
```
vlasenko: Should be instead:
```
export default async () => new Promise(resolve => knex.transaction(resolve));
```

modules/user/server-ts/resolvers.js
```diff
@@ -88,8 +94,17 @@ export default pubsub => ({
 
           e.throwIf();
 
-          const [createdUserId] = await User.register({ ...input });
-          await User.editUserProfile({ id: createdUserId, ...input });
+          const passwordHash = await createPasswordHash(input.password);
+
+          const trx = await createTransaction();
+          let createdUserId;
+          try {
```
vlasenko: I think all the code inside this resolver should be in a transactional `try` `catch` block

vlasenko: I'm not 100% sure

evgenijolshanskij: @vlasenko Also, `const trx = await createTransaction();` cannot be wrapped in `try /catch` since `trx` won't be available in the `catch` block;

StKostyantin: Add certificate in a transaction is complete

vlasenko:
Looks good to me. Thank you @StKostyantin!

Issue #771, opened 2018-07-12
Support for pdf reporting and importing data to Excel


coooolers:

mitjade:
I think this might be outside the scope of what this kit should provide. I needed both of this in one of my projects and ended up using http://pdfmake.org/ and https://sheetjs.com.

vlasenko:
@mitjade It is not out of the scope of what kit should provide if the solution will use open source technologies. These tasks are pretty generic.

vlasenko:
Implemented in #985

PR #985, opened 2018-12-28
Support pdf reporting 

StKostyantin:
This pull request add pdf reporting. Also, it fixes the #771 issue.


modules/reports/client-react/components/ReportPreview.jsx
```diff
@@ -0,0 +1,74 @@
...
+  </tr>
+);
+
+ContactRow.propTypes = {
+  values: PropTypes.array
+};
+
+const ContactHeader = ({ headerItem }) => (
+  <Fragment>
+    {headerItem.map((item, key) => (
```
alekseyzadorozhniy: Use plural variable instead of singular

modules/reports/client-react/locales/ru/translations.json
```diff
@@ -0,0 +1,14 @@
+{
+  "navLink": "Репорты",
+  "title": "Репорты в пдф",
+  "list": {
+    "title": {
+      "front": "Репорт с фронта",
+      "server": "Репорт с сервера"
+    }
+  },
+  "btn": "Распечатать репорт",
```
alekseyzadorozhniy: Could you choose one version of the translation?
For me "отчет" is better

modules/reports/client-react/containers/Report.jsx
```diff
@@ -0,0 +1,65 @@
...
+    const { t } = this.props;
+    const button = <Button>{t('btn')}</Button>;
+
+    return (
+      <PageLayout>
+        {this.renderMetaData()}
+        <Query query={ReportQuery}>
+          {({ loading, error, data }) => {
+            if (loading) return 'Loading...';
+            if (error) return `Error! ${error.message}`;
```
alekseyzadorozhniy: We do not handle errors in this way. Look at similar components

modules/reports/client-react/containers/Report.jsx
```diff
@@ -0,0 +1,65 @@
...
+  render() {
+    const { t } = this.props;
+    const button = <Button>{t('btn')}</Button>;
+
+    return (
+      <PageLayout>
+        {this.renderMetaData()}
+        <Query query={ReportQuery}>
+          {({ loading, error, data }) => {
+            if (loading) return 'Loading...';
```
alekseyzadorozhniy: All texts which are shown to the user should be translated

modules/reports/server-ts/index.ts
```diff
@@ -0,0 +1,15 @@
+import ServerModule from '@module/module-server-ts';
+
+import Report from './sql';
+import schema from './schema.graphql';
+import createResolvers from './resolvers';
+import pdfResolvers from './generateFile/pdf/resolvers';
```
evgenijolshanskij: `pdf` and `excel` should come as separate modules. Also, they shouldn't be located in the `generateFile` folder, it is redundant.

modules/reports/client-react/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Report from './containers/Report';
+import resources from './locales';
+
+const NavLinkWithI18n = translate('report')(({ t }: { t: TranslateFunction }) => (
+  <NavLink to="/report" className="nav-link" activeClassName="active">
+    {t('navLink')}
+  </NavLink>
+));
+
+export default new ClientModule({
```
evgenijolshanskij: The Report `ClientModule` should have the same structure the server has. That means it should be divided into several modules: `excel`, `pdf`, and each of them provide its part of functionality. Currently all the logic is blended.

vlasenko:
Thank you @StKostyantin and @Sasha-belokur. Awesome work!

modules/reports/client-react/pdf/client/components/Button.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Button as RSButton } from 'reactstrap';
+
+interface ButtonProps {
+  children: any;
+}
+
+export default class Button extends React.Component<ButtonProps> {
```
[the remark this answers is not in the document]
vlasenko: Good point @Werter12 I missed that

modules/reports/client-react/pdf/client/containers/WithExportPDF.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { Fragment, Component } from 'react';
+import ReactToPrint from 'react-to-print';
```
BaloPhilip: why use 'react-to-print' library, can use native browser functionality 'window.print', the same result.

modules/reports/server-ts/excel/helpers/generateExcel.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+};
+const style = wb.createStyle({
+  font: {
+    color: '#000000',
+    size: 12
+  }
+});
+
+export default function generateBufferExcel(reports: object[]) {
```
lyzhovnik: Please write interface for reports prop.

modules/reports/server-ts/excel/resolvers.ts
```diff
@@ -0,0 +1,10 @@
+import generateExcel from './helpers/generateExcel';
+
+export default () => ({
+  Query: {
+    async excel(obj: any, arg: any, { Report }: any) {
```
lyzhovnik: Type for Report DAO

modules/reports/server-ts/pdf/helpers/PDFBuilder.ts
```diff
@@ -0,0 +1,56 @@
+import PdfPrinter from 'pdfmake';
+import fonts from './fonts/Roboto/';
+
+export default class PDFBuilder {
+  private printer: any;
```
lyzhovnik: Types

modules/reports/server-ts/pdf/helpers/PDFBuilder.ts
```diff
@@ -0,0 +1,56 @@
...
+      style,
+      alignment
+    });
+  }
+
+  public addStyle(name: string, style: object) {
+    this.styles[name] = style;
+  }
+
+  public addTable(data: object[], columnsWidth: object) {
```
lyzhovnik: Types for props instead of `object`

modules/reports/server-ts/pdf/resolvers.ts
```diff
@@ -0,0 +1,11 @@
+import generator from './helpers/generator';
+
+export default () => ({
+  Query: {
+    async pdf(obj: any, arg: any, { Report, req }: any) {
```
lyzhovnik: Types for `Report` and `req`

modules/reports/server-ts/sql.ts
```diff
@@ -0,0 +1,10 @@
+import { knex } from '@gqlapp/database-server-ts';
+
+export default class Report {
+  public report() {
```
lyzhovnik: It looks like you selecting all data, then it should have name `reports` or even `users` because you selecting users from `report` table. @evgenijolshanskij  @vlasenko  what do you think about this?

vlasenko: @lyzhovnik Should be named `getUsers` here I think

evgenijolshanskij: The DB table name should reflect the data it contains. Despite the name of the module, the table and DAO class are needed to be renamed (`contacts`, `users` etc.).

evgenijolshanskij: I'd use different from `user` name for not confusing anyone. Just make up some other entity.

modules/reports/server-ts/excel/helpers/generateExcel.ts
Sasha-belokur: Fixed in #1023

modules/reports/server-ts/excel/resolvers.ts
Sasha-belokur: Fixed in #1023

modules/reports/server-ts/pdf/helpers/PDFBuilder.ts
Sasha-belokur: Partially Fixed in #1023, printer is third-party library, so we can't add types for it.

modules/reports/server-ts/pdf/helpers/PDFBuilder.ts
Sasha-belokur: Fixed in #1023

modules/reports/server-ts/pdf/resolvers.ts
Sasha-belokur: Fixed in #1023

modules/reports/server-ts/sql.ts
Sasha-belokur: Fixed in #1023

modules/reports/server-ts/pdf/helpers/PDFBuilder.ts
vlasenko: @Sasha-belokur Why? How about:
https://www.npmjs.com/package/@types/pdfmake

Sasha-belokur: I added types from package above. But I couldn't add type for server printer object since these types are for front-end library. We discussed with @lyzhovnik, that it would be better to get rid of constructor and assign properties on declaration phase.

Issue #692, opened 2018-04-05
Add New User Not Working


coooolers:
When I clcik on Add the URL is set to: http://localhost:3000/users/0
The id should be set to 4.

I had a look at the code, it looks like the system is not taking care of the Trimming before saving the username, email address,and the user's First and Last Name. Trim all these fields before a record is saved to the database.

When I try to set the Active checkbox, it does nothing.

When I try to save the User, a click on the Save button does nothing.

Thanks

coooolers:
Add 4 blank spaces in the Username field and it happily accepts the input without taking caring of the white spaces.

coooolers:
As a side note: Trimming should also be taken care of when a new post is created.

mitjade:
Id is set to 0 to indicate a new user is being added. This way edit form can be reused for adding and editing.

Regarding trimming. This should be an easy fix `str.trim()`. Could be added in `packages/client/src/modules/user/components/UserForm.web.jsx` -> `handleSubmit`. Fill free to submit a PR to fix this.

abhishek-kumar-91:
I want to work on this issue.

Rutwic15:
I am facing problem while setting up the project locally. While installing yarn it is showing many errors. I want to work on this issue. Please help me

abhishek-kumar-91:
> I am facing problem while setting up the project locally. While installing yarn it is showing many errors. I want to work on this issue. Please help me

Ok, I will help you to setup project on locally.

Agniv25:
I want to work on this. Can you please assign this to me.

ATIQ-SHAIKH:
I would like to work on this issue, can someone assign me this please.

PR #993, opened 2019-01-10
Add validation method

basilalex:
This PR fixes bug with a falsy value in the username field described in [#692](https://github.com/sysgears/apollo-universal-starter-kit/issues/692) issue.


vlasenko:
Why do we do trimming inside validation? Validation - validates correctness of the fields it must not transform them

vlasenko:
@basilalex We have validation method that covers this and similar cases: `match`. No need to implement all the variations that can already be handled by `match`.

In this specific case - the proper approach is to `trim` user input, but not inside `validation` package - but rather in a formik form handling code

PR #994, opened 2019-01-10
Copyedit locales

AndSviat:
Fixed various spelling errors and mistakes in the translations for all the server and client modules.


AndSviat:
Fixed the tested messages in `.spec.js` files so that tests now pass successfully for copyedited locales.

Issue #752, opened 2018-06-13
Add Firebase authentication


vlasenko:
Add authentication via Firebase

adarshanand67:
Is this issue still open? I would like to work on it.
I'm a first timer.
Could you put me through on the details so I could know what to do?

Gautam-10137:
Hii Is this issue still open? I would like to work on it.

divyamojas:
Hey, is this issue still open? 
Assign me, I will do it.

Ansh1372:
Hey this issue id still available , I  would like to work on it.

aasimtaif:
Hi is this issue still open I would like to work on this I have created the same feature in this project https://github.com/aasimtaif/Instagram-clone

Aarav238:
i would like to work on it. Please assign this to me . @larixer

bilalshaikh292:
Heyy the issue have not been solved yet ? I would like to contribute ? is the issue still open ?

PR #996, opened 2019-01-11
Add Firebase auth

Iwasherd:
This PR adds Firebase auth  issue #752
Also create Firebase module


packages/server/firebase-admin-sdk-data.json
```diff
@@ -0,0 +1,12 @@
+{
```
vlasenko: This file should not be placed into Git I think due to security reasons

PR #1001, opened 2019-01-14
Add descriptions for new Yarn scripts

AndSviat:
Updated the `docs/yarnScripts.md` doc: added descriptions for three new commands that run only an iOS, an Android, or a web application.


vlasenko:
Thank you @AndSviat, looks good!

PR #1022, opened 2019-02-01
Fix I18n

BaloPhilip:
If remove any locales from langList config I18n,  the application still displayed all locales which are in  folder 'locales'

### DIFF
--- a/modules/look/client-react/ui-antd/components/LanguagePicker.jsx
+++ b/modules/look/client-react/ui-antd/components/LanguagePicker.jsx
@@ -12,12 +12,19 @@ export default class LanguagePicker extends React.Component {
   );
 
   render() {
+    const { i18n } = this.props;
+    const languages = Object.keys(i18n.store.data);
+
     return (
-      <Dropdown overlay={this.getMenuItems()}>
-        <a className="ant-dropdown-link" href="#">
-          {this.props.i18n.language.slice(0, 2).toUpperCase()} <Icon type="down" />
-        </a>
-      </Dropdown>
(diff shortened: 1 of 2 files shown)


modules/look/client-react/ui-antd/components/LanguagePicker.jsx
```diff
@@ -12,12 +12,19 @@ export default class LanguagePicker extends React.Component {
...
+    const languages = Object.keys(i18n.store.data);
+
     return (
-      <Dropdown overlay={this.getMenuItems()}>
-        <a className="ant-dropdown-link" href="#">
-          {this.props.i18n.language.slice(0, 2).toUpperCase()} <Icon type="down" />
-        </a>
-      </Dropdown>
+      <React.Fragment>
+        {languages.length > 1 && (
```
vlasenko: What if languages.length === 1 ?

vlasenko: Ah, nevermind, then we don't need to choose :)

vlasenko:
LGTM, thank you!

modules/look/client-react/ui-antd/components/LanguagePicker.jsx
evgenijolshanskij: @vlasenko, In this case, we don't need the language picker to be displayed.

vlasenko: @evgenijolshanskij Yes, exactly!

Issue #1027, opened 2019-02-09
Update all the modules to use React Hooks where appropriate


vlasenko:
We should use React Hooks everywhere where it makes sense over class components

vlasenko:
Implemented in #1029

PR #1029, opened 2019-02-12
Update all the modules to use React Hooks where appropriate #1027

BaloPhilip:


vlasenko:
Thank you @BaloPhilip. Awesome work!

PR #1041, opened 2019-03-05
Make sign up flow more user-friendly. Issue #1039

BaloPhilip:
Issue #1039

PR #1042, opened 2019-03-06
Fixed import statement in User module

beingtmk:
Profile component was imported instead of container


vlasenko:
@beingtmk Thanks. Should rather be `import Profile ...`. These change is done as part of #1041

PR #1056, opened 2019-04-04
Fix problem of serving all static files as index.html when SSR is disabled

seanjsong:
Since we are using req.path.indexOf('.') < 0 to check if the server is serving static files such as *.jpg *.css *.js , when SSR is disabled we still need to check it, otherwise the server will be responding with the content of index.html to any static file requests.


vlasenko:
@seanjsong It is actually correct for SPA server to respond on each request with `index.html` and let `SPA` client-side logic to handle URL

seanjsong:
When I mean is this:

Browser requests for favicon.ico,  server responds 200, seems correct, but its content is actually the content of index.html.

This PR is trying to fix this problem.

vlasenko:
Thank you for explanations, looks good!

PR #1057, opened 2019-04-09
Create common naming for React, React Native

BaloPhilip:


BaloPhilip:
changed approach to working with the stack of technology

PR #1059, opened 2019-04-09
Remove the extra padding on top of every screen in iOS app

seanjsong:
Before fix:
before

After fix:
after

Please have a look at [this issue](https://github.com/react-navigation/react-navigation/pull/3633) to get an idea of why we need to fix in this way. Basically:

* react-native has improved the implementation of ``<SafeAreaView>``, but react-navigation has hardcoded ``forceInset`` prop value which no longer works correctly
* someone noted this hardcoded ``forceInset`` prop value caused unremovable padding on top, and suggested it should be removed
* the maintainer of react-navigation prefers making this prop customizable while preserving its hardcoded value: https://github.com/react-navigation/react-navigation-stack/blob/master/src/views/Header/Header.js#L605

That's why we need to customize the ``forceInset`` prop (effectively removing it).

Hopefully anyone who starts his/her own project based on this codebase would no longer have to figure out first thing how to remove the padding on top :)


vlasenko:
Great insight, thank you very much Sean!

PR #1060, opened 2019-04-16
Move SSR and GraphQL middlewares to separate modules

basilalex:
This PR is moving React SSR middleware and GraphQL middleware to separate modules.


modules/core/server-ts/app.ts
```diff
@@ -1,39 +1,30 @@
-import express from 'express';
+import express, { Express } from 'express';
 import path from 'path';
 import { GraphQLSchema } from 'graphql';
+import ServerModule, { Ware, CreateGraphQLContext } from '@gqlapp/module-server-ts';
```
vlasenko: What is the meaning of `Ware` ? Why `Ware` ?

modules/core/server-ts/app.ts
```diff
@@ -1,39 +1,30 @@
...
 
-import { isApiExternal } from '@gqlapp/core-common';
-import ServerModule from '@gqlapp/module-server-ts';
-
-import graphiqlMiddleware from './middleware/graphiql';
-import websiteMiddleware from './middleware/website';
-import createApolloServer from './graphql';
 import errorMiddleware from './middleware/error';
 
+type TApplyWare = (ware: Ware) => void;
```
vlasenko: Why `TApplyWare`, why do you add `T` here, did you see this practice anywhere in the kit?

PR #1062, opened 2019-04-17
Move GraphQL setup to the separate module

basilalex:
This PR will move GraphQL setup to the separate module.


modules/module/server-ts/ServerModule.ts
```diff
@@ -5,6 +5,7 @@ import { Express } from 'express';
 import { ConnectionParamsOptions } from 'subscriptions-transport-ws';
 import { IResolvers } from 'graphql-tools';
 import CommonModule, { CommonModuleShape } from '@gqlapp/module-common';
+import { GraphQLSchema } from 'graphql';
```
evgenijolshanskij: All the GraphQL-related code/logic/exports should be located in the newly created `graphql` module.

modules/core/server-ts/app.ts
```diff
@@ -1,39 +1,32 @@
...
 import errorMiddleware from './middleware/error';
 
+type ApplyMiddleWare = (fn: MiddlewareFunc) => void;
+
 export const createServerApp = (schema: GraphQLSchema, modules: ServerModule) => {
   const app = express();
   // Don't rate limit heroku
   app.enable('trust proxy');
+  const { appContext, beforeware, middleware } = modules;
+  const createGraphQLContext: CreateGraphQLContext = (req, res) => modules.createContext(req, res);
```
evgenijolshanskij: Why can't we move creating PgaphQL context to the `graphql` module?

modules/module/server-ts/ServerModule.ts
basilalex: Done

modules/graphql/server-ts/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const middleware: MiddlewareFunc = (app, appContext, { createGraphQLContext, schema }) => {
+  app.get('/graphiql', graphiqlMiddleware);
+
+  if (!isApiExternal) {
+    const graphqlServer = createApolloServer({ createGraphQLContext, schema });
+    graphqlServer.applyMiddleware({ app, path: __API_URL__, cors: { credentials: true, origin: true } });
+  }
+};
+
+const makeGQLContextCreator: MakeGQLContextCreator = modules => (req, res) => modules.createContext(req, res);
```
evgenijolshanskij: `createContext` function shouldn't belong to the `ServerModule` anymore. Instead, the new `GraphQLModule` should be created that provides this function.

evgenijolshanskij: The same is for `schema` and `resolvers` params of the `ServerModule`

vlasenko: @evgenijolshanskij Currently `GraphQLModule` is a client-side GraphQL feature modules interface

evgenijolshanskij: @vlasenko updated the module name.

modules/graphql/server-ts/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@gqlapp/graphql-types-server-ts",
+  "name": "@gqlapp/graphql-server-ts",
```
vlasenko: Should we really merge `graphql-types` and `graphql` modules together? 

The other consideration is the name of the module `graphql` it is too ambigious if we just want to move GraphQL server-side setup here. Maybe we should name module `graphql-server` then?

evgenijolshanskij: @vlasenko personally I think that it is quite reasonable. I don't think we will need graphql types without using gpaphql in general. Also, it allows keeping things in one place which is more comfortable.

modules/graphql/server-ts/index.ts
vlasenko: @evgenijolshanskij Yes, sounds good to me, also we can remove client-side `GraphQLModule` to `GraphQLClientModule`, we will not break backwards compatibility by doing this

modules/graphql/server-ts/package.json
vlasenko: @evgenijolshanskij I don't have strong opinion here, so I'm okay with this merge. How about `graphql` name? Right now we have server setup there, so it should be named `graphql-server` maybe?

basilalex: The name would look like "@gqlapp/graphql-server-server-ts" because of the name convention if we rename the folder.

vlasenko: @basilalex yes

modules/graphql/server-ts/index.ts
basilalex: Done.

modules/graphql/server-ts/createApolloServer.ts
```diff
@@ -0,0 +1,37 @@
...
+  const context = async ({ req, res }: { req: any; res: any }) => ({
+    ...(await createGraphQLContext(req, res)),
+    req,
+    res
+  });
+
+  const formatError = (error: ApolloError) =>
+    error.message === 'Not Authenticated!' ? new AuthenticationError(error.message) : error;
+
+  const formatResponseCustom = (response: any, options: { [key: string]: any }) =>
```
vlasenko: If you want to stress that this is a custom function, its name should be `customFormatResponse`. But still it will be the bad name, because it does not reveal what this function does. A better name here is `formatAndLogResponse`

modules/graphql/server-ts/types.ts
```diff
@@ -0,0 +1,71 @@
...
+import ServerModule from '@gqlapp/module-server-ts';
+
+/**
+ * A function to make a GraphQL context creator
+ *
+ * @param modules App modules
+ *
+ * @returns function that creates a GraphQL context
+ */
+export type MakeGQLContextCreator = (modules: ServerModule) => (req: Request, res: Response) => any;
```
vlasenko: Types must be declared closer to the locations where they are used. It is a very bad practice to use `MakeGQLContextCreator` in only a single place and move declaration of the type to another place.

modules/graphql/server-ts/package.json
basilalex: Done.

modules/graphql/server-ts/createApolloServer.ts
basilalex: Renamed.

modules/graphql/server-ts/types.ts
basilalex: Fixed.

vlasenko: @basilalex What about all the other types?

modules/module/server-ts/contracts/ServerModule.ts
```diff
@@ -0,0 +1,42 @@
+import { Express } from 'express';
+import GraphQLServerModule, { GraphQLServerModuleShape } from './GraphQLServerModule';
+
+/**
+ * A function which registers new middleware.
+ *
+ * @param app an instance of Express
+ * @param appContext application context
+ * @param options application additional options
```
vlasenko: What is `application additional options`. This is totally unclear

Issue #763, opened 2018-07-01
Code: Ability to switch ORM to Sequelize or Mongoose


lgants:
Does anyone know of a fork of this project that uses Sequelize or Mongoose? I love absolutely everything about this project with one big exception: knex. Knex itself lacks support for more advanced relations (i.e. belongs to many). This functionality can be added with Bookshelf, but neither knex nor Bookshelf library to seem to be very actively maintained (as of when I created this post last merge for both was 3 months ago; compared with 5 days ago for Sequelize and 4 days  ago for Mongoose). I've created a fork with Sequelize, which is getting close, but I've become stuck: https://github.com/lgants/apollo-universal-starter-kit-sequelize. Is it possible to add support for another ORM?

vlasenko:
@lgants knex is well maintained. The last commit was 21 hours ago. The question is how to maintain both Sequelize and knex in the kit? How Sequilize can be added to the kit to minimize maintenance cost? And who will maintain Sequilze-related code in the kit?

lgants:
@vlasenko Those are good questions. I'd like to contribute to the project. Unfortunately, in the process of implementing Sequelize in my fork it's become clear that it'd be difficult to maintain both. What are your thoughts on including [BookshelfJS](http://bookshelfjs.org/) instead? That would provide support for polymorphic associations. I could look into the implementation.

vlasenko:
@lgants Yes, bookshelfJS looks better. @mairh @mairh Ujjwal, Mitja, what do you think guys?

mitjade:
@vlasenko I have looked at bookshelfJS before and do see some advantages, but ended up using just knex, since it offers more flexibility with what I needed.

vlasenko:
@mitjade Can you use bookshelfJS for 90% code base and just go down to knex in 10% when you need flexibility. Is this possible?

mitjade:
@vlasenko It’s been a while since I looked at this, so I am not sure right now, can take a quick look tomorrow.

lgants:
@vlasenko it's probably worth noting that the BookshelfJS build is failing (and has been for at least a few days). There's also a notice in the docs about the library undergoing some "changes." I found an interesting article about JS ORMs/ODMs that was fairly critical of Bookshelf: https://medium.freecodecamp.org/a-comparison-of-the-top-orms-for-2018-19c4feeaa5f. 

[...]

PR #1063, opened 2019-04-23
Started adding ObjectionJS with Counter Module

beingtmk:
Refer Issue: Support for other ORMs such as Sequelize or Mongoose #763

### DIFF
--- a/modules/counter/server-ts/sql.ts
+++ b/modules/counter/server-ts/sql.ts
@@ -1,11 +1,31 @@
 import knex from '../../../packages/server/src/sql/connector';
+import { Model } from 'objection';
 
-export default class Counter {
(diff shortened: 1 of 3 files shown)


modules/counter/server-ts/sql.ts
```diff
@@ -1,11 +1,31 @@
 import knex from '../../../packages/server/src/sql/connector';
+import { Model } from 'objection';
+
+// Give the knex instance to objection.
+Model.knex(knex);
```
vlasenko: Do we really need to call it here? This is app-global call, I guess, so it should be somewhere in @gqlapp/database-server-ts module.

Also what about migrations and seeds for Counter?

PR #1080, opened 2019-05-15
bring i18next-express-middleware into effect

seanjsong:
Currently i18next-express-middleware doesn't do its work. Server doesn't return the correct message in client's locale. This commit fixes the problem.

Before fix:

Screen Shot 2019-05-13 at 12 56 25 PM

After fix:

Screen Shot 2019-05-15 at 2 15 28 PM


seanjsong:
Note:

The emails sent by server should also be in the same locale as corresponding client request. I noticed email subjects and bodies are hardcoded in English, such as [this](https://github.com/sysgears/apollo-universal-starter-kit/blob/3f18b700722821b04d6644ac53d754272ab9a279/modules/user/server-ts/password/resolvers.js#L86).

With i18next-express-middleware in effect this can also be fixed. However I don't speak Russian so I can't help with translating these email subjects and bodies.

vlasenko:
@seanjsong Cool, many thanks! How about adding translations in your native language? :)

PR #1081, opened 2019-05-15
make mobile app send API request with locale parameter

seanjsong:
A further fix following up with [the last commit](https://github.com/sysgears/apollo-universal-starter-kit/commit/274ef8200a76d588a97f54fe20dfd672f732d9bd).

Problem:
Web client can send locale information to server through cookie, but mobile app can't. Although locale information is stored in Expo SecureStore in [LanguageDetector.native.ts](https://github.com/sysgears/apollo-universal-starter-kit/blob/master/modules/i18n/common-react/LanguageDetector.native.ts), API requests sent by mobile app won't include this locale information.

There is a way to manipulate cookies in mobile app, namely [react-native-cookies](https://github.com/joeferraro/react-native-cookies), but it's a native package and requires ejecting from Expo.

My solution is passing locale information as query string in API requests.

Before fix:
Screen Shot 2019-05-15 at 8 53 08 PM

After fix:
Screen Shot 2019-05-15 at 9 00 56 PM

I'm only suggesting one possible solution, not very neat, nor very modular (coupling i18n module with core module). You can merge this PR or choose to fix this problem in some other way.

### DIFF
--- a/modules/core/common/createApolloClient.ts
+++ b/modules/core/common/createApolloClient.ts
@@ -12,10 +12,12 @@ import ApolloCacheRouter from 'apollo-cache-router';
 import { hasDirectives } from 'apollo-utilities';
 import { DocumentNode } from 'graphql';
 import { IResolvers } from 'graphql-tools';
+import i18n from 'i18next';
 
 import settings from '@gqlapp/config';
 
 import log from './log';
+import { PLATFORM } from '../../../packages/common/utils';
 
 interface CreateApolloClientOptions {
   apiUrl?: string;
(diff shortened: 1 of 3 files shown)


seanjsong:
Besides, there are still more locale related problems to be fixed:

* when users click the link in their "email confirmation" or "password reset" email, they will be directed to a web page, that page should also be rendered in the same locale as their email.

* when users reset their password on the web page, they will receive another confirmation email, that email should also be in the same locale.

vlasenko:
@seanjsong Maybe we should accept locale on the server from cookie, if it is not preset in cookie, then from header, if it is not present in header, then from HTTP url query parameter? After that we will be able to implement all the cases that you mentioned and keep the code decoupled. For example for React Native it will then make sense to pass locale in a header. In links inside email we will be able to use HTTP url querry param

PR #1085, opened 2019-05-20
Rename docs for consistency and update links accordingly

AndSviat:
Renamed all the documents under `docs` to use "More Readable Names" instead of "camelCaseNames" and updated the links in `README.md` and all the docs accordingly.


vlasenko:
Thank you @AndSviat, looks good!

PR #1088, opened 2019-05-24
1006 fb native flow

ChingxWando:
There is a problem with expo not reading my process.env variables, it reads settings and /config/auth.js alright but I had to hard copy facebook app/ client ID in auth.js to make it work, just a quick FYI if anyone is testing this code.


vlasenko:
@ChingxWando It is correct, that it doesn't read `.env`. `.env` contains sensitive information to be used on the server only.

vlasenko:
`.env` must contain server-only and secret settings. `FACEBOOK_CLIENTID` is neither server-only, nor secret, it must not be in `.env` at all.

PR #1099, opened 2019-06-10
Convert React Native modules to use React Hooks

BaloPhilip:


packages/mobile/.zenrc.js
```diff
@@ -7,7 +7,7 @@ const config = {
       defines: {
         __CLIENT__: true
       },
-      enabled: false
+      enabled: true
```
vlasenko: You can just use `yarn watch:android` and not override this option by hand

PR #1100, opened 2019-06-11
Switch to webpack.config.js for building artifacts

vlasenko:
This PR switches from `@larix/zen` to `webpack.config.js` for building React frontend and React backend.


eric-burel:
Hi, what's the future of Zen in this case?

Currently my pain point is debugging, because Zen obfuscate the underlying webpack config, so it makes it hard to debug a new feature like supporting Storybook without getting your help. But there seems to be some espace hatch to handle the config.

In Vulcan we have similar questionnings about the build system, which is currently one of the last feature that makes us dependent to Meteor, since it's one of the only framework to support client + server + mobile build without any config. However Meteor can't support Storybook and we had to write custom webpack loaders to run it despite of this limitation.

vlasenko:
@eric-burel Zen future is not decided yet. Perhaps Zen will be reworked to generate default webpack configs and write them to disk and you will be able to see them and extend, I thought about generating `webpack.base.config.js` using Zen (which you will be able to update by running new version of Zen), and your custom code to extend that config might be in `webpack.config.js`. As for the kit, I want to switch to `webpack.config.js` first everywhere and test this concept together with introducing code splitting support

eric-burel:
Yes generating files instead of dynamic structures is also something I am thinking a lot about. It may end up being easier to manage for the developers because it gives full control while making things easier. 

In Vulcan we automate a lot of things (generating the Mongo database structure, the gql schema and the resolvers/mutations) but sometimes I'd be glad that it output a Mongoose Schema or a GraphQL file instead of just dynamically define them, because it would make it very easy to tweak it or to pass it to another development tool.

From the framework standpoint it may be way harder to code though due to the lack of advanced serialization tool, so you can't convert a JS object into a JS file easily. So you have to write a converter "by hand" and manipulate templates a lot, or use tools like AST but why not.

PR #1110, opened 2019-06-17
Update docs, remove zen, make relative URLs

vlasenko:
This PR removes Zen mentions and uses relative URLs for documentation links

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ great possibilities for styling for your web and mobile applications.
 * Fractal modular architecture that's easy to support and extend
 * The possibility to create modules in TypeScript _and_ JavaScript at the same time
 * No need to develop [typical features](#architecture-and-implemented-modules) for your applications
-* Zero project configuration thanks to [Zen], a builder for JavaScript projects
 
 #### I am a Project Manager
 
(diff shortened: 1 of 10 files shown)


strefethen:
@vlasenko I'd definitely be interested in better understanding why zen was removed? There are a lot of changes in the commit and the above comment doesn't offer enough context to understand the rational. Seems like this is a pretty big change that's not well explained.

vlasenko:
@strefethen Why zen was removed. Zen is a tool that checks project's dependencies in `package.json` and tries to generate `good` default Webpack config which can further be customized. This PR actually takes Webpack config generated by Zen and stores the config into kit source code itself. What has been dropped is config autogeneration. Config autogeneration has obvious pros that users don't have to spend time to write and maintain "default good" webpack config. But there is also a price to pay when you need more flexibility with tweaking webpack config. Yes, `zen` uses `webpack-merge` and different strategies to make config tweaking very flexible, though there are cases that they are not flexible enough though.  For example try to add code splitting support into zen generated webpack config. Or try to have completely different strategy with using mini-css-webpack-plugin. The changes like these will require essential changes to webpack config.

So the main rationale is to provide more means to tweak webpack config as much as you want and as easy as possible.

PR #1111, opened 2019-06-18
Support custom namespaces for modules

verdverm:
add support for custom npm namespace in server, it crashes unless you get your own into that regex

where else might this be an issue?

- `packages/client/webpack.config.js` there is a node_modules exclude thing in there that has the `@gqlapp`

basically anywhere `@gqlapp` shows up, we would want to add support for adding in the `NPM_WHITELIST` env var I used in the first commit.

btw, is there another name you would prefer for that ENV_VAR @vlasenko ?
- NPM_NAMESPACES ?

### DIFF
--- a/docs/Importing Modules.md
+++ b/docs/Importing Modules.md
@@ -45,3 +45,24 @@ can use it by importing necessary classes or components with ES6 `import`.
 
 [yarn documentation]: https://yarnpkg.com/lang/en/docs/workspaces/
 [webpack dependency management]: https://webpack.js.org/guides/dependency-management/
+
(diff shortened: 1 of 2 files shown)


vlasenko:
@verdverm I would name it something like: `MODULENAME_REGEX` and droped 2 lines about it into the docs

verdverm:
@vlasenko I changed the logic to match the regex name, such that it becomes the complete regex, and if not specified, falls back to the original

vlasenko:
@verdverm Thank you very much!

PR #1112, opened 2019-06-18
Extract-files breaking changes

eric-burel:
Now Extract-files handles the cloning and returns an object `{clone: <clone of your variables>, files: Map }` that contains the files.


vlasenko:
Many thanks Eric!

PR #1119, opened 2019-06-28
Redirect unauthenticated users in AuthRoute

eric-burel:
Previously AuthRoute would not redirect unauthenticated users.


vlasenko:
Looks good, many thanks Eric!

PR #1120, opened 2019-07-01
Fix regression from #1119

eric-burel:
Redirects correctly logged in user for pages with "reversed" permission (reset password, log in, password forgotten etc.)


vlasenko:
Thanks!

PR #1130, opened 2019-07-27
switch to using nodemon-webpack-plugin

verdverm:
The current implementation will stop restarting the server after so many file changes (non-deterministic)

This switches to using https://www.npmjs.com/package/nodemon-webpack-plugin and seems to be more stable.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -168,6 +168,7 @@
     "jsdom": "^11.5.1",
     "mkdirp": "^0.5.1",
     "nodemon": "^1.19.1",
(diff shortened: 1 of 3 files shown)


vlasenko:
@verdverm Tony, many thanks for this PR! I'm working on a webpack plugin that will be specially crafted for Node backend launching under Webpack HMR:
https://github.com/sysgears/node-hmr-plugin

vlasenko:
See #1133 and 
https://github.com/sysgears/node-hmr-plugin

PR #1142, opened 2019-08-21
Read field meta data for nested form data, too

theWeird:
Enables FieldAdapter to determine touched and error state not only for the top level form object, but for nested fields, too.

Imagine a Post with some attached files created in the same form.  
That could create a form values object like `{title: 'foo', files: [{filename: 'bar.txt'}]`

Whilst formik.touched['title'] works good, formik.touched['files[0].filename'] doesn't.  
Using the lodash get method, the nested path can be easily extracted from formiks meta objects, too.


vlasenko:
Cool, many thanks @theWeird !

PR #1143, opened 2019-08-22
Bug fix: The Pagination demo doesn't work and continuously reload 

vadimderkach:
Bugfix: The Pagination demo doesn't work and continuously reload because the dependency array is missed in the useEffect hook of useDataProvider.

### DIFF
--- a/modules/pagination/client-react/containers/DataProvider.jsx
+++ b/modules/pagination/client-react/containers/DataProvider.jsx
@@ -38,7 +38,7 @@ export const useDataProvider = () => {
 
   useEffect(() => {
     loadData(0, 'replace');
-  });
+  }, [loadData]);
 
   return { items, loadData };
 };
--- a/modules/pagination/client-react/containers/DataProvider.jsx
+++ b/modules/pagination/client-react/containers/DataProvider.jsx
@@ -35,10 +35,10 @@ export const useDataProvider = () => {
     },
     [items]
   );
-
-  useEffect(() => {
+  const useMountEffect = fn => useEffect(fn, []);
+  useMountEffect(() => {
     loadData(0, 'replace');
-  }, [loadData]);
+  });
 
   return { items, loadData };
 };


modules/pagination/client-react/containers/DataProvider.jsx
```diff
@@ -35,8 +35,8 @@ export const useDataProvider = () => {
     },
     [items]
   );
-
-  useEffect(() => {
+  const useMountEffect = fn => useEffect(fn, []);
```
vlasenko: Thanks, but why do you create `useMountEffect` here, why just not pass empty array to `useEffect`?

vadimderkach: ESLint eslint-plugin-react-hooks doesn't allow passing an empty dependency array. It autocorrects it. See https://github.com/facebook/react/issues/15063

vlasenko: In this case lets maybe use `// eslint-disable-next-line ...` to disable the eslint rule for this line, since we know this is a valid use case and not the case where we just forgot deps. There is no deps, it should be fired one time.

vadimderkach: IMHO disabling ESLint rule doesn't clearly demonstrate the developer's intent. It doesn't have a clear explanation or a hint "why". By declaring useMountEffect we are adding a semantic to it (clearly stating that we would like to use it only on mount), and avoiding ESLint rule disabling which is not desirable.

vlasenko: Well, declaring useMountEffect is not much better solution as well, since we don't solve the underlying problem, but rather mask it, by introducing additional function wrapper. The underlying problem is that the render code itself is not written cleanly and it should be rewritten instead, IMHO.

vlasenko:
@vadimderkach Thank you. I'm merging this now for bugfix purposes, but better fix would still be needed - e.g. rewriting pagination code

PR #1148, opened 2019-08-27
Correct separator when generating package name

mitjade:
Use correct separator in module name when generating package name.


vlasenko:
Cool, thank you Mitja!

PR #1155, opened 2019-10-08
Redirect back to original url, after redirected to login and logged in.

mitjade:

If you access a protected url, you get redirected to login and after you login you would want to navigate to initial url.

**How did you fix it?**

Added an `redirectBack` parameter that gets set when you get redirected to login and then use it to navigate to it, if login is successful.


modules/core/server-ts/middleware/website.tsx
```diff
@@ -106,7 +106,13 @@ const renderServerSide = async (req: any, res: any, schema: GraphQLSchema, modul
   context.pageNotFound === true ? res.status(404) : res.status(200);
 
   if (context.url) {
-    res.writeHead(307, { Location: context.url });
+    const redirectBack =
+      context.location && context.location.state && context.location.state.from && context.location.state.from.search
```
vlasenko: @mitjade Why do we check `context.location.state.from.search` here?

vlasenko: Can we simplify this logic if we will have on client-side: `{ from: props.location.pathname }`, e.g. pass in URL from client-side, not object?

mitjade: If you access `kitUrl/users` it first redirects you to `.../profile` and then to `login`. So I needed to pass the param through multiple redirects.

vlasenko: @mitjade Got you, thanks. But I still think passing `{ from: props.location.search || props.location.pathname  }` from client will be less confusing then passing whole object to the server and let server handle it.

mitjade: @vlasenko I took this naming convention from https://reacttraining.com/react-router/web/example/auth-workflow. Will try and just provide a string. Maybe then we could have some naming collisions in this case?

mitjade: @vlasenko It looks like I have to use `state: { from: props.location }`, to be able to access it in `website.tsx`, otherwise it does not work.

vlasenko: @mitjade Yes, but you can pass in URL directly from the client-side code, so that you don't need to do processing on the server, e.g. pass from client:
`state: { from: props.location.search || props.location.pathname }`

modules/user/client-react/containers/Auth.web.jsx
```diff
@@ -7,19 +7,29 @@ import { withLoadedUser } from './AuthBase';
 const AuthRoute = withLoadedUser(
   ({ currentUser, role, redirect = '/login', redirectOnLoggedIn, component: Component, ...rest }) => {
     const RenderComponent = props => {
+      const redirectBack = props.location.search
```
vlasenko: It would be slightly simpler to read if it were:
`const redirectBack = props.location.search || (props.location.pathname ? `?redirectBack=${props.location.pathname}` : '')`

mitjade: @vlasenko Now if you try accessing `/users` and get redirected to `/login?redirectBack=/users`. Then login as `user`, you get redirected back from `/users` do to insufficient access rights to `/`, you are left with `?redirectBack=/users`.

mitjade: Before `redirectBack` was added only when you got redirected on the server, now this gets added also if you click on links you do not have access to.

vlasenko: @mitjade I'm not sure I follow. I have tried to reproduce this problem on your PR. I enter url with `/users` at the end as unauthenticated user, then I login and the I get redirected to `/users` just fine. Could you clarify the problem and tell me how to reproduce it with current code in your PR ?

mitjade: Try to login as `user@example.com` not `admin@example.com`.

vlasenko: @mitjade The problem is that we don't distinguish between the cases when user is not logged in and when the user is logged in, but role doesn't match.
We have:
```js
      // redirect users unlogged or without sufficient permissions
      return isRoleMatch(role, currentUser) ? (
        <Component currentUser={currentUser} {...props} {...rest} />
      ) : (
        <Redirect to={{ pathname: redirect }} />
      );
```
But we should rather have:
```js
      if (currentUser) {
        return isRoleMatch(role, currentUser) ? (
          <Component currentUser={currentUser} {...props} {...rest} />
        ) : (
          // If user is logged in, but has no permissions to access URL redirect him permanently, without an option to redirect back
          <Redirect to={redirect} />
        );
      } else {
        // If user is not logged in, remember where we should redirect him back
        return <Redirect to={redirect + redirectBack} />;
      }

```

PR #1156, opened 2019-10-09
Explicitly specify returning column when inserting a user

seanjsong:

I ran into an exception when this line of code is executed:

https://github.com/sysgears/apollo-universal-starter-kit/blob/969cc51155d7be88f97e27f305bf678836fa9220/modules/user/server-ts/social/facebook/index.js#L23

Then I found out nothing was returned because my db happens to be postgresql, as stated in knexjs documents: 

```
// Returns [1] in "mysql", "sqlite", "oracle"; [] in "postgresql" unless the 'returning' parameter is set.
knex('books').insert({title: 'Slaughterhouse Five'})
```

...

**How did you fix it?**

Explicitly specify the returning column i.e. 'id' column with returning()

...

### DIFF
--- a/modules/user/server-ts/sql.js
+++ b/modules/user/server-ts/sql.js
@@ -138,7 +138,9 @@ class User {
   }
 
   register({ username, email, role = 'user', isActive }, passwordHash) {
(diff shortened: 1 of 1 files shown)


vlasenko:
Many thanks @seanjsong !

PR #1164, opened 2019-11-13
Fix no session on social Logins

MassivDash:
Because social tokens are constructed via await access.grantAccess(user, req, user.passwordHash), and password_hash was missing session could not be established.

### DIFF
--- a/modules/user/server-ts/social/shared.js
+++ b/modules/user/server-ts/social/shared.js
@@ -1,4 +1,5 @@
 import { access } from '@gqlapp/authentication-server-ts';
+import bcrypt from 'bcryptjs';
 import User from '../sql';
 
 export async function onAuthenticationSuccess(req, res) {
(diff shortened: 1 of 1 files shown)


modules/user/server-ts/social/shared.js
```diff
@@ -14,10 +15,12 @@ export async function onAuthenticationSuccess(req, res) {
 }
 
 export const registerUser = async ({ id, username, displayName, emails: [{ value }] }) => {
+  const passwordHash = await bcrypt.hash(id || username || displayName, 12);
```
larixer: Why do we use id or username or displayName as a password? Why do we use 12 as a salt?

MassivDash: I have replicated the brcypt password hash creation from the access modules, it had 12 as salt. Since social user passport strategy has no user password filed, we need to create whatever password in order for the social tokens to be created. I thought might as well create the password from user outside id, like it was before but if Is is missing the  create one from display name. My assumption is that either user will continue to login through social or he will request forgot password.

larixer: I don't think we need a password at all for social users, because this field is not used as a password. I think the better approach is to handle the case elsewhere in the code that password might be actually missing: #1165

PR #1166, opened 2019-11-27
Make find command in README BSD/macOS compatible

Ithamar:

The find command in the README does not work on BSDs / macOS.
...

**How did you fix it?**

Add a directory specification to the find command (.).

...


larixer:
Thank you!

PR #1168, opened 2019-12-06
fix subscription pages being inside head

realalexhomer:
In the web subscription module the `Auth` and `AddSubscription` containers are rendered inside of a `Helmet` component. This is causing the pages to render inside of the `<head>` tags of the page.

There isn't an error message being thrown but the pages are blank when you navigate to them. To reproduce follow the steps for stripe set up and navigate to the subscription pages.
...

**How did you fix it?**
Moved the contents of the pages outside of `Helmet`
...


larixer:
LGTM, thanks!

PR #1171, opened 2019-12-14
Converted Counter module to use ObjectionJS

beingtmk:
#763 

**Note:** Reading from the docs of ObjectionJS, it is not a full ORM but just a relational query builder, hence it doesn't support creating migrations, we will have to use knex for that. seeds might be possible, but surfing online I have not found anybody seeding with ObjectionJS


larixer:
@beingtmk Thank you, looks great!

PR #1179, opened 2019-12-23
Ant Design RenderSelect Issue

lgants:
The RenderSelect for Ant Design on master uses the native HTML element. Consequently, it isn't styled with Ant Design.

current_master_2019-12-22 18-54-30 2019-12-22 18_54_43

Reactstrap [Form](https://reactstrap.github.io/components/form/) Input component with type `select` onChange receives an event. However, Ant Design [Select](https://ant.design/components/select/) component onChange receives a string value. 

...

**How did you fix it?**
To avoid modifying behavior in FieldAdapter, which is agnostic of UI library (excluding mobile vs non-mobile), I've passed an onChange func that provides a mocked event to the formik handleChange function. Thoughts?

after_2019-12-22 18-44-47 2019-12-22 18_45_38

This avoids any impact on the existing reactstrap Input component.
bootstrap_after_2019-12-22 18-47-19 2019-12-22 18_47_38

...

### DIFF
--- a/modules/forms/client-react/FieldAdapter.jsx
+++ b/modules/forms/client-react/FieldAdapter.jsx
@@ -26,11 +26,12 @@ class FieldAdapter extends Component {
   }
 
   onChange = e => {
-    const { onChange } = this.props;
+    const { formik, onChange } = this.props;
+
     if (onChange) {
       onChange(e.target.value, e);
     } else {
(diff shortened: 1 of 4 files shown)


larixer:
@lgants Cool, looks great, many thanks!

PR #1180, opened 2019-12-24
Fix antd right menu for AntD

mitjade:

Ant Design `MenuItem` has problems rendering correctly if it is nested inside another `div`.

Screenshot 2019-12-24 at 09 56 29

**How did you fix it?**

Removed extracted component, since it is not used anywhere else.

Screenshot 2019-12-24 at 09 55 29


larixer:
Thank you Mitja!

PR #1185, opened 2020-02-18
Correction for Angular frontend execute command

sateesh2020:
The command for `Angular frontend execute` is wrongly typed


larixer:
Thank you @sateesh2020 !

PR #1188, opened 2020-03-14
Add EMAIL_SENDER for mailer

justinr1234:

The user that you login to your SMTP provider with is not always the same as the `from` address that you’d like to use when mail is sent to users.

**How did you fix it?**

Added new environmental variable for `EMAIL_SENDER` with fallback to `EMAIL_USER`


larixer:
@justinr1234 Thanks! LGTM

PR #1189, opened 2020-03-14
Fix mysql SSL issue

justinr1234:

The mySQL driver doesn't like string values for the `DB_SSL` environmental key.

**How did you fix it?**

Make sure to convert the value from `"true"` or `"false"` only and pass it onto the config properly. Don't set the `ssl` param unless the environmental variable is specifically set.


justinr1234:
Nevermind

PR #1190, opened 2020-03-14
Add loading assets from react

justinr1234:

The ability to access assets from other files inside React components.

**How did you fix it?**

* Exported the resolved paths from the `favicon` module so that I can access the logo
* Added logo to assets

Future work could be a more generic way of exporting assets shared across multiple modules, which would imply making this code more generic.

### DIFF
--- a/config/app.js
+++ b/config/app.js
@@ -1,5 +1,6 @@
 export default {
   name: 'Apollo Starter Kit',
+  logo: 'logo.svg', // Loaded via webpack and stored in favicon/common/assets
   logging: {
     level: (process.env.npm_config_argv || '').search(/(watch|start)/) >= 0 ? 'debug' : 'info',
     debugSQL: false,
--- a/docs/Importing Modules.md
+++ b/docs/Importing Modules.md
@@ -30,12 +30,28 @@ folder:
 
 ```js
 // modules/favicon/common/index.js
-// Favicon.ico should not be hashed, since some browsers expect it to be exactly on /favicon.ico URL
-require('!file-loader?name=[name].[ext]!./assets/favicon.ico'); // eslint-disable-line
+const exportedAssets = {};
 
-// Require all files from assets dir recursively adding them into assets.json
-let req = require.context('!file-loader?name=[hash].[ext]!./assets', true, /.*/);
-req.keys().map(req);
+if (!__TEST__) {
+  // Favicon.ico should not be hashed, since some browsers expect it to be exactly on /favicon.ico URL
+  require('!file-loader?name=[name].[ext]!./assets/favicon.ico'); // eslint-disable-line
+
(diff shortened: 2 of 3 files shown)


config/app.js
```diff
@@ -1,5 +1,6 @@
 export default {
   name: 'Apollo Starter Kit',
+  logo: 'logo.svg', // Loaded via webpack and stored in favicon/common/assets
```
larixer: I don't see where does this property used in this PR?

docs/Importing Modules.md
```diff
@@ -30,12 +30,28 @@ folder:
...
-let req = require.context('!file-loader?name=[hash].[ext]!./assets', true, /.*/);
-req.keys().map(req);
+if (!__TEST__) {
+  // Favicon.ico should not be hashed, since some browsers expect it to be exactly on /favicon.ico URL
+  require('!file-loader?name=[name].[ext]!./assets/favicon.ico'); // eslint-disable-line
+
+  // Require all files from assets dir recursively addding them into assets.json
+  let req = require.context('!file-loader?name=[hash].[ext]!./assets', true, /.*/);
+  req.keys().map(key => {
+    exportedAssets[`${key.replace('./', '')}`] = req(key);
```
larixer: I'm not sure this is enough though, maybe also use `humps.camelize` ?

modules/favicon/common/index.js
```diff
@@ -1,8 +1,14 @@
...
+
 if (!__TEST__) {
   // Favicon.ico should not be hashed, since some browsers expect it to be exactly on /favicon.ico URL
   require('!file-loader?name=[name].[ext]!./assets/favicon.ico'); // eslint-disable-line
 
   // Require all files from assets dir recursively addding them into assets.json
   let req = require.context('!file-loader?name=[hash].[ext]!./assets', true, /.*/);
-  req.keys().map(req);
+  req.keys().map(key => {
+    exportedAssets[`${key.replace('./', '')}`] = req(key);
```
larixer: I'm not sure this is enough though, maybe also use `humps.camelize` ?

config/app.js
justinr1234: It isn’t used, it’s optional for someone to use it elsewhere. I put it in the docs.

modules/favicon/common/index.js
justinr1234: @larixer I don’t understand what you’re asking. The purpose of this line is to allow you to reference the assets by name later on. It is meant to be equivalent to how you can access assets on the server (notably in the file `website.tsx`.

The files are all prefixed with `./` by default ai this just removes that. Otherwise you just reference each file by whatever name it is on disk.

larixer:
Thanks, LGTM

PR #1212, opened 2021-04-16
Fix tests

vsevolodbvd:

Client-side tests are broken now

**How did you fix it?**

Fix tests and warnings


larixer:
Awesome, many thanks Seva!

Issue #1113, opened 2019-06-20
Docs refer to user.js which appears to be in auth.js and has been somewhat restructured


strefethen:
**Describe the bug**
The docs refer to "config/user.js" which does not exist and the structure of this file is not consistent with config/auth.js. There are several differences with the organization of auth.js that are not documented.

**To Reproduce**
1. View this search: https://github.com/sysgears/apollo-universal-starter-kit/search?q=%22user.js%22&unscoped_q=%22user.js%22

**Expected behavior**
Notice there are results in the docs that refer to "user.js" which does not exist. These settings are available in auth.js and the structure of the file is somewhat different than what's described.

The expectation is that the documentation reflect the current state of user/auth configuration.

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser: Brave
 - Version:  Version 0.65.121

PR #1235, opened 2023-12-14
Made changes to docs/FAQ and docs/Configuration.

Minus0149:

As shown in issue #1113 "Docs refer to user.js which appears to be in auth.js and has been somewhat restructured" 

**How did you fix it?**

Changed the document according to the instructions given as the original was pointing towards wrong file.


larixer:
Thank you, looks good!

Minus0149:
You are welcome. Thank you for the opportunity.

